PR #9, opened 2019-04-09
Updating superdesk core

akintolga:


petrjasek:
can you update also client to match?

PR #14, opened 2019-07-09
New fields filename and stt-topic added

akintolga:


server/stt/parser.py
```diff
@@ -67,5 +67,27 @@ def set_extra_fields(self, item, xml):
...
+        # filename
+        try:
+            link_node = xml.find(self.qname('itemMeta')).find(self.qname('link'))
+
+            if link_node:
+                filename = link_node.find(self.qname('filename')).text
+                if filename:
+                    item.setdefault('extra', {})['filename'] = filename
+
+        except Exception:
```
petrjasek: can you use something more specific?

PR #32, opened 2022-11-24
[STTNHUB-174] Ingest Event contact information

MarkLark86:
Depends on
* https://github.com/petrjasek/eve-elastic/pull/101
* https://github.com/superdesk/superdesk-core/pull/2404

### DIFF
--- a/server/stt/stt_events_ml.py
+++ b/server/stt/stt_events_ml.py
@@ -33,15 +33,29 @@ def search_existing_contacts(contact: Dict[str, Any]) -> Optional[Dict[str, Any]
     if contact.get("first_name") and contact.get("last_name"):
         first_name = contact["first_name"].lower()
         last_name = contact["last_name"].lower()
+
         cursor = contacts_service.search({
             "query": {
                 "bool": {
-                    "must": [{
-                        "query_string": {
-                            "query": f'first_name:"{first_name}" AND last_name:"{last_name}"'
-                        }
-                    }]
-                }
+                    "must": [
+                        {
+                            "match": {
+                                "first_name": {
+                                    "query": first_name.lower(),
+                                    "operator": "AND",
+                                },
+                            },
+                        },
+                        {
+                            "match": {
+                                "last_name": {
+                                    "query": last_name.lower(),
+                                    "operator": "AND",
+                                },
+                            },
+                        },
+                    ],
+                },
             },
             "sort": ["_score"]
         })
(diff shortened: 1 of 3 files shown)


server/stt/stt_events_ml.py
```diff
@@ -1,12 +1,45 @@
...
+}
+
+
+def search_existing_contacts(contact: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+    """Attempt to find existing media contact using email, falling back to first_name/last_name combo"""
+
+    contacts_service = get_resource_service("contacts")
+    if len(contact.get("contact_email") or []):
+        cursor = contacts_service.search(
+            {"query": {"bool": {"must": [{"term": {"contact_email.keyword": contact["contact_email"][0]}}]}}}
```
petrjasek, line 28: it might be useful to call `.lower()` on the email as it won't match if we get the email from ingest with some uppercase

server/stt/stt_events_ml.py
```diff
@@ -1,12 +1,45 @@
...
+        )
+        if cursor.count():
+            return list(cursor)[0]
+
+    if contact.get("first_name") and contact.get("last_name"):
+        cursor = contacts_service.search({"query": {"bool": {"must": [
+            {"term": {"first_name.keyword": contact["first_name"]}},
```
petrjasek: same here regarding `.lower()`, with the name it's probably more important

server/stt/stt_events_ml.py
```diff
@@ -31,10 +31,20 @@ def search_existing_contacts(contact: Dict[str, Any]) -> Optional[Dict[str, Any]
...
-            {"term": {"first_name.keyword": contact["first_name"]}},
-            {"term": {"last_name.keyword": contact["last_name"]}}
-        ]}}})
+        first_name = contact["first_name"].lower()
+        last_name = contact["last_name"].lower()
+        cursor = contacts_service.search({
+            "query": {
+                "bool": {
+                    "must": [{
+                        "query_string": {
```
petrjasek: might work better with match:
```
must: [
    {"match": {"first_name": first_name}},
    {"match": {"last_name": last_name}},
]
```
so there won't be any formatting issues

petrjasek (approved):

MarkLark86:
@petrjasek Needing to use the expanded version of `match`, for first/last names that have multiple. `match` uses `OR` as the default `operator`

petrjasek (approved):

PR #37, opened 2023-06-29
Showing STT’s priority (importance) on planning items [STTNHUB-200]

devketanpro:


server/tests/stt_planning_ml_test.py
```diff
@@ -13,11 +13,26 @@ def test_stt_metadata(self):
...
-        self.assertEqual(self.item["coverages"][0]["coverage_id"], "ID_WORKREQUEST_159799")
+        self.assertEqual(
+            self.item["coverages"][0]["coverage_id"], "ID_WORKREQUEST_159799"
+        )
+
+        # Urgency [STTNHUB-200]
+        self.assertIn(
+            {
+                "name": "Keskipitkä juttu (medium length)",
+                "qcode": "stturgency:3",
```
petrjasek: lets make the qcode like `stturgency-3` so it's not using `:` which might make it split later in the formatter

server/data/vocabularies.json
```diff
@@ -251,5 +251,31 @@
         "service": {"all": 1},
         "init_version": 1,
         "items": []
+    },
+    {
+        "_id": "importance",
```
petrjasek: it would be better to match the `scheme`, actually in general we don't put the scheme in the items, it needs to be there only when using the items and adding it to the article

petrjasek: so this should be actually `stturgency`

server/tests/stt_planning_ml_test.py
devketanpro: yep

petrjasek (approved):

PR #39, opened 2023-07-07
remove translation text from 'importance' cv and refactor code [STTNHUB-200]

devketanpro:


server/data/vocabularies.json
```diff
@@ -261,19 +261,19 @@
         "selection_type": "single selection",
```
MarkLark86, line 261: Need to increment `init_version` attribute, otherwise the changes won't be applied automatically on deploy

devketanpro: yep

MarkLark86 (approved):

PR #42, opened 2023-10-19
fix : Enabling urls without http in front of them [STTNHUB-252]

devketanpro:


server/stt/common.py
```diff
@@ -6,27 +6,57 @@
...
+            original[config.ID_FIELD], original
+        )
     elif original.get("pubstatus") != POST_STATE.CANCELLED:
-        update_post_item({"pubstatus": POST_STATE.CANCELLED, "_etag": original["_etag"]}, original)
+        update_post_item(
+            {"pubstatus": POST_STATE.CANCELLED, "_etag": original["_etag"]}, original
+        )
+
+
+def transform_link_from_text(item, fields):
```
MarkLark86, line 45: We already have this code in `superdesk-core` (https://github.com/superdesk/superdesk-core/blob/develop/superdesk/text_utils.py#L193).

This is the code that is currently being used to transform plain text to rich text (for Editor3 fields).
Seems like the bug is there, and should be fixed there

server/stt/stt_events_ml.py
```diff
@@ -86,6 +104,20 @@ def parse(self, tree: Element, provider=None):
                 # If the item contains the ``sttinstruct:remove`` signal, no need to ingest this one
                 continue
             self.set_extra_fields(item, tree)
+
+            fields = (
```
MarkLark86, line 108: This functionality is already in EventsML parser from `superdesk-planning`. And we have to check if the field is using `Editor3`, otherwise adding HTML tags in plain text would not be ideal.

As we support this as a core feature, we should update the EventsML and PlanningML parsers to support all possible fields. See https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/feed_parsers/events_ml.py#L252 for example.

We first check if the field is Editor3 in the content profile, and then attempt to convert to html. So it might be good to implement a utility in `superdesk-planning` to convert multiple fields on ingest to editor3, checking it's content profile.

MarkLark86 (changes requested):

MarkLark86:
@devketanpro Basically this PR can be closed, and a PR in `superdesk-core` and `superdesk-planning` so this can be supported as a core improvement & bug fix

server/stt/common.py
devketanpro: I see. Okay, let me check. Thanks, Mark!

PR #43, opened 2023-11-24
[STTNHUB-292] fix: Use empty slugline for Placeholder Coverages

MarkLark86:


devketanpro:
hi, @MarkLark86 , I wonder: if we remove the slugline from these placeholder coverages, how can the user identify whether this is a placeholder coverage or not?

MarkLark86:
> hi, @MarkLark86 , I wonder: if we remove the slugline from these placeholder coverages, how can the user identify whether this is a placeholder coverage or not?

From a logical point of view (from code) we have the placeholder flag.

The customer doesn't want this slugline shown in Newshub (especially considering it's in English).

And, at this point in time, the customer doesn't use Superdesk directly. It's more of a headless CMS for them

devketanpro (approved):
make sense; thanks for clarifying.

PR #47, opened 2024-02-02
[SDESK-7166] fix(ingest): Skip linking coverages that already have an assignment

MarkLark86:
Also made the hook code more robust by catching & logging errors, allowing ingested item to succeed ingest procedure

Depends on https://github.com/superdesk/superdesk-planning/pull/1906


server/stt/signal_hooks.py
```diff
@@ -141,7 +166,13 @@ def before_content_published(_sender: Any, item: Dict[str, Any], updates: Dict[s
         logger.warning(f"Failed to get 'assignment_id' of coverage '{coverage_id}'")
         return
 
-    _link_assignment_and_content(assignment_id, coverage_id, item.get("guid"), True)
+    try:
+        _link_assignment_and_content(assignment_id, coverage_id, item.get("guid"), True)
+    except Exception as err:
+        logger.exception(err)
+        logger.error("Failed to link coverage assignment to content")
```
petrjasek, line 173: would be good in general to combine those to avoid an extra report on sentry, something like:
```py
logger.exception("Failed to link coverage assignment to content (%s)", err)
```
should log all the info into single report

petrjasek (approved):

PR #91, opened 2024-08-22
Fix server tests

petrjasek:


server/features/ingest_events.feature
```diff
@@ -18,7 +18,7 @@ Feature: Ingest STT Event Items
         """
         {"_items": [{
             "guid": "urn:newsml:stt.fi:259431",
-            "state": "spiked"
+            "state": "killed"
```
MarkLark86, line 21: When did this change? https://github.com/superdesk/superdesk-stt/blob/develop/server/stt/common.py#L20
If the Event has not been published, and is in `ingested` state, it should be spiked

petrjasek:
only one I can think of now is https://github.com/superdesk/superdesk-planning/pull/2009

server/features/ingest_events.feature
petrjasek: only one I can think of now is this one https://github.com/superdesk/superdesk-planning/pull/2009

MarkLark86 (approved):

PR #101, opened 2024-09-23
[STT-81]: Update STT EventsML ingest parser: add organisation to contacts 

msilven:
Changed stt_events_ml.py to get organization data and add it into the contact data.

### DIFF
--- a/server/stt/stt_events_ml.py
+++ b/server/stt/stt_events_ml.py
@@ -229,6 +229,7 @@ class STTEventsMLParser(EventsMLParser):
             last_name = contact_info.find(self.qname("lastname", ns=NS["stt"]))
             job_title = contact_info.find(self.qname("title", ns=NS["stt"]))
             phone = contact_info.find(self.qname("phone"))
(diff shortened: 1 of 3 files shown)


MarkLark86 (commented):
@msilven Some tests would be good to make sure this always works.
[server/tests/events_ml_parser_test.py](https://github.com/superdesk/superdesk-stt/blob/develop/server/tests/events_ml_parser_test.py) and [fixtures](https://github.com/superdesk/superdesk-stt/tree/develop/server/tests/fixtures)

msilven:
> @msilven Some tests would be good to make sure this always works. [server/tests/events_ml_parser_test.py](https://github.com/superdesk/superdesk-stt/blob/develop/server/tests/events_ml_parser_test.py) and [fixtures](https://github.com/superdesk/superdesk-stt/tree/develop/server/tests/fixtures)

There should be tests now for this. Also I missed flake8, nosetests and behave earlier. Those are now installed too and I made sure all tests are ok before commit/push.

MarkLark86 (approved):

PR #178, opened 2025-02-28
Feature/stt spellchecker

0is1:
Lingosoft proofread feature for STT.

https://trello.com/c/wUSmm5E1


server/stt/spellcheckers/stt_fin.py
```diff
@@ -0,0 +1,167 @@
...
+                "text": text,
+                "language": "fi",
+                "domain": "Standard",
+                "options": {
+                    "Format": "json",
+                },
+            }
+            # Add API key to header "apiKey"
+            headers = {"apiKey": self.api_key, "Content-Type": "application/json"}
+            r = requests.post(
```
petrjasek, line 52: might be better to use session if it will be used more often

petrjasek (approved):

PR #203, opened 2025-05-05
Modify the SD locations handling - to store the location data received from the STT (event) ingests [STT-100]

devketanpro:


petrjasek:
would it automatically store those locations in locations db?

devketanpro:
> would it automatically store those locations in locations db?

No, it won’t automatically store those locations in the locations DB. I’ll create a separate PR to update the parser logic accordingly.

server/settings.py
```diff
@@ -115,3 +115,50 @@ def env(variable, fallback_value=None):
     "a",
     "pre",
 )
+
+SCHEMA_UPDATE = {
```
petrjasek: I wonder how does it affect the system?

devketanpro: we already have core functionality to update the schema dynamically. you can see the implementation here:
🔗 [superdesk-core/resource.py#L177–L186](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/resource.py#L177-L186)

So most likely, we’ll just need to trigger an Elastic reindex once the schema is updated.

petrjasek: I rather wonder what will happen, atm those stt fields won't be stored?

devketanpro: the ticket mentions that we need to update the Elasticsearch mapping — that’s why I updated the schema to enable indexing in Elasticsearch.
Without this schema update, the fields are still stored in the database.

MarkLark86:
Why is this PR marked with `Do not merge`?

devketanpro:
> Why is this PR marked with `Do not merge`?

Hi @MarkLark86,

Just to clarify so we can move this PR forward — the main open question is whether we do need to update the Elasticsearch mapping for the STT location fields, or if that part can be skipped for now.

Currently, the fields are stored in the database

PR #205, opened 2025-05-06
Modify the Parser to store the location data received from the STT (event) ingests [STT-100]

devketanpro:
Schema PR :- https://github.com/superdesk/superdesk-stt/pull/203

### DIFF
--- a/server/stt/stt_events_ml.py
+++ b/server/stt/stt_events_ml.py
@@ -280,18 +280,29 @@ class STTEventsMLParser(EventsMLParser):
         locations_service = get_resource_service("locations")
         if locations_service is not None:
             try:
-                existing_location = locations_service.find_one(
-                    req=None,
-                    name=location.get("name"),
-                    **{"address.city": location.get("address", {}).get("city")},
+                stt_id = (
+                    location.get("address", {}).get("extra", {}).get("sttlocationalias")
                 )
-                if existing_location:
-                    location_id = existing_location["_id"]
-                else:
-                    location_id = locations_service.post([location])[0]
+                custom_guid = f"urn:stt:location:{stt_id}" if stt_id else None
 
-                # Store both the full location data and reference ID
-                item["location"] = [{**location, "location_id": location_id}]
+                existing_location = (
+                    locations_service.find_one(req=None, guid=custom_guid)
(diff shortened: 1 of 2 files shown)


server/tests/events_ml_parser_test.py
```diff
@@ -187,3 +187,41 @@ def test_search_contacts_case_insensitive(self):
...
+        self.location_data = {
+            "name": "Sokos Hotel Presidentti",
+            "address": {
+                "title": "Sokos Hotel Presidentti",
+                "city": "Helsinki",
+                "state": "Uusimaa",
+                "country": "Suomi",
+                "line": ["Eteläinen Rautatiekatu 4"],
+                "extra": {
+                    "sttlocationalias": "14068",
```
petrjasek, line 203: would probably use that number to generate id so we don't rely on name when looking it up later, eg. generating something like `urn:stt:location:14068` as location id

server/stt/stt_events_ml.py
```diff
@@ -280,18 +280,29 @@ def set_location_details(self, item, location_xml, notes):
...
         if locations_service is not None:
             try:
-                existing_location = locations_service.find_one(
-                    req=None,
-                    name=location.get("name"),
-                    **{"address.city": location.get("address", {}).get("city")},
+                stt_id = (
+                    location.get("address", {}).get("extra", {}).get("sttlocationalias")
                 )
+                custom_guid = f"urn:stt:location:{stt_id}" if stt_id else None
```
petrjasek: we would need a guid, so if there is no stt_id I wouldn't store it probably and just keep it on the event

server/stt/stt_events_ml.py
```diff
@@ -276,7 +276,35 @@ def set_location_details(self, item, location_xml, notes):
...
+                )
+
+                if stt_id:
+                    custom_guid = f"urn:stt:location:{stt_id}"
+                    existing_location = locations_service.find_one(
+                        req=None, guid=custom_guid
+                    )
+
+                    if existing_location:
+                        saved_location = existing_location
```
petrjasek: it might be useful to update the location data if it exists

petrjasek (approved):

PR #219, opened 2025-05-20
Modify the SD locations handling - to store the location data received from the STT (event) ingests [STT-100]

devketanpro:
Fix : unique_name is missing in the location data

### DIFF
--- a/server/stt/stt_events_ml.py
+++ b/server/stt/stt_events_ml.py
@@ -238,9 +238,7 @@ class STTEventsMLParser(EventsMLParser):
                 if country_el is not None:
                     country_name = country_el.find(self.qname("name"))
                     country = country_name.text if country_name is not None else None
-            location["unique_name"] = (
-                f"{name}, {city}, {country}" if (city and country) else name
-            )
+            location["unique_name"] = ", ".join(filter(bool, [name, city, country]))
             location["address"]["title"] = name
         except AttributeError:
             pass
--- a/server/stt/stt_events_ml.py
+++ b/server/stt/stt_events_ml.py
@@ -217,6 +217,7 @@ class STTEventsMLParser(EventsMLParser):
         try:
             name = location_xml.find(self.qname("name")).text
             location["name"] = name
+            location["unique_name"] = name
             location["address"]["title"] = name
         except AttributeError:
             pass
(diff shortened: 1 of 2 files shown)


server/stt/stt_events_ml.py
```diff
@@ -217,6 +217,7 @@ def set_location_details(self, item, location_xml, notes):
         try:
             name = location_xml.find(self.qname("name")).text
             location["name"] = name
+            location["unique_name"] = name
```
petrjasek: not sure if that would be unique actually, wouldn't it require also city and country or something?

devketanpro: you're right name alone isn't unique. Let's include city and country.

server/stt/stt_events_ml.py
```diff
@@ -217,7 +217,18 @@ def set_location_details(self, item, location_xml, notes):
         try:
             name = location_xml.find(self.qname("name")).text
             location["name"] = name
-            location["unique_name"] = name
+            address_xml = location_xml.find(self.qname("POIDetails")).find(
```
petrjasek: this is also used later and seems like it might be empty, then this whole part would fail and there would be no `unique_name`

server/stt/stt_events_ml.py
```diff
@@ -217,7 +217,18 @@ def set_location_details(self, item, location_xml, notes):
...
+                self.qname("address")
+            )
+            city = (
```
petrjasek: not sure if there will be always city, but if not this would block the branch

server/stt/stt_events_ml.py
```diff
@@ -215,8 +215,32 @@ def set_location_details(self, item, location_xml, notes):
...
+                if locality is not None:
+                    city_name = locality.find(self.qname("name"))
+                    city = city_name.text if city_name is not None else None
+
+                country_el = address_xml.find(self.qname("country"))
+                if country_el is not None:
+                    country_name = country_el.find(self.qname("name"))
+                    country = country_name.text if country_name is not None else None
+            location["unique_name"] = (
+                f"{name}, {city}, {country}" if (city and country) else name
```
petrjasek: might be better to do it like:
```
", ".join(filter(bool, [name, city,  country]))
```
so it can handle missing city

petrjasek (approved):

PR #265, opened 2025-06-18
update client 20250618

tomaskikutis:


petrjasek:
I didn't cherry pick that livequery fix in planning extension, will do it now

PR #298, opened 2025-07-03
Npm Update

dzonidoo:
link: https://github.com/superdesk/superdesk-planning/pull/2391


petrjasek:
better do it via dependabot, we don't want to switch from node 22 to node 14 here

PR #315, opened 2025-07-17
Updated package-lock.json

BrianMwangi21:


petrjasek:
we use node 22 for it now @BrianMwangi21 , or you can also trigger the update via depenedabot

PR #346, opened 2025-08-11
implement task STT-1219 lottery veikkaus

minhpham212:
# implement task [STT-1219]  lottery veikkaus

## Summary

Adds VeikkausTextFeedParser to handle Veikkaus plain-text and legacy XML-wrapped lottery feeds, converting them into a single Superdesk ingest item.

## Key Points
	•	Detects encoding with chardet + fallbacks, fixes common Finnish mojibake.
	•	Supports .txt, .xml, and <root><p>…</p></root> legacy format.
	•	Formats content into <p> with <br/> separators.
	•	Generates guid from filename, sets headline from first non-empty line or defaults to "Veikkaus".
	•	Adds metadata: extra.veikkaus (department, desk, filename), keywords ["Veikkaus", "lottery"].

## Benefit

Standardizes ingestion of Veikkaus lottery results and ensures clean, correctly encoded text.

[STT-1219]: https://sofab.atlassian.net/browse/STT-1219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/.DS_Store
petrjasek, line 1: pls remove that file

server/requirements.txt
```diff
@@ -347,3 +347,4 @@ wrapt==1.17.2
     # via elastic-apm
 xmlsec==1.3.14
     # via superdesk-core
+chardet==5.2.0
```
petrjasek: add it also to `requirements.in` otherwise it will get lost on update

server/tests/stt_parse_veikkaus_test.py
```diff
@@ -0,0 +1,377 @@
...
+    def test_fixture_file_parsing(self):
+        """Test parsing the actual fixture file."""
+        items = self.parser.parse(self.fixture_path)
+        self.assertEqual(len(items), 1)
+        item = items[0]
+
+        # Test that headline is first line of fixture
+        self.assertEqual(item["headline"], "TOTO75-ravi")
+        self.assertIn("<p>", item["body_html"])
+        self.assertEqual(item["description_text"], "pelitulos.Y_06.08.2022.txt")
```
petrjasek: not sure if that will be useful, might make it to the output

server/requirements.txt
minhpham212: I removed chardet from requirements.txt.

server/.DS_Store
minhpham212: Can you suggest a way to write test cases for this function?

server/requirements.txt
minhpham212: I removed chardet in requirements.txt

server/tests/stt_parse_veikkaus_test.py
minhpham212: Can you suggest a way to write this test?

petrjasek: the test is fine, I was wondering if it's useful to put `pelitulos.Y_06.08.2022.txt` into description, it might make it to the output

server/requirements.txt
petrjasek: imo it should be in both `requirements.txt` and `requirements.in`, you can add it to `requirements.in` only and then use [pip-tools](https://github.com/jazzband/pip-tools?tab=readme-ov-file#requirements-from-requirementsin) to update `requirements.txt`

petrjasek (commented):
pls add a test for `versioncreated` to be be populated

server/.DS_Store
petrjasek, line 1: pls remove this file, fyi you can add it to your global gitignore

server/tests/stt_parse_veikkaus_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+from tests import TestCase
+from stt.stt_parse_lottery_veikkaus import (
+    VeikkausTextFeedParser,
+    to_body_html,
+    fix_encoding_issues,
+)
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
```
petrjasek, line 14: there is no use for that in tests probably

minhpham212: updated

server/.DS_Store
minhpham212: updated

minhpham212:

.gitignore
```diff
@@ -47,3 +47,5 @@ include/
 client/index.html
 client/templates-cache.js
 client/npm-debug.log
+server/.DS_Store
+server/.coveragerc
```
petrjasek: those shouldn't be here as it's not project specific, you can set your [global ignored files locally](https://stackoverflow.com/questions/7335420/global-git-ignore)

server/stt/stt_parse_lottery_veikkaus.py
```diff
@@ -0,0 +1,165 @@
...
+        item: Dict[str, Any] = {
+            "guid": guid,
+            "type": "text",
+            "headline": headline,
+            "body_html": body_html,
+            "description_text": filename,
+            "original_source": "STT",
+            "urgency": 4,
+            "pubstatus": "usable",
+            "versioncreated": datetime.utcnow().isoformat() + "Z",
```
petrjasek: this should be actually `datetime` instance and not string

petrjasek (approved):

PR #352, opened 2025-08-15
[STT-1217] implement feature Feeding service for Superdesk content API

minhpham212:
## [STT-1217] implement feature Feeding service for Superdesk content API

1. add  file erver/stt/io/feed_parsers/stt_parse_content_api
2. add file server/stt/io/feed_parsers/stt_parse_content_api.py
3. add file server/tests/stt_content_api_test.py
4. add file server/tests/stt_parse_content_api_test.py
### Configurable Provider Fields setting:
1. url
2. api_key

[STT-1217]: https://sofab.atlassian.net/browse/STT-1217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #353, opened 2025-08-17
implement task [STT-1218] Feeding service for TT API

minhpham212:
## Implement Task Feature: STT-1218 - Feeding Service for TT Content API

Feeding Service: STTContentAPIService – Handles API communication and data fetching

Parser: ContentAPITTItemParser – Transforms TT API responses into a Superdesk-compatible format

## Add unit tests for the following files:
	1.	server/tests/stt_tt_content_api_test.py
	2.	server/tests/stt_tt_parse_content_api_test.py


petrjasek:
it's referring to content api in a few places (in code and PR name) but it's not related, would be good to rename it to just tt api to avoid confusion. also there is some code duplicated from https://github.com/superdesk/superdesk-stt/pull/352, would be good to merge that one first and then reuse it

minhpham212:
> it's referring to content api in a few places (in code and PR name) but it's not related, would be good to rename it to just tt api to avoid confusion. also there is some code duplicated from #352, would be good to merge that one first and then reuse it

Because the API returns two different data structures, we need two separate services to parse them accordingly.

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,140 @@
...
+                processed[tf] = self._normalize_timestamp(processed[tf])
+
+        # Ensure versioncreated exists and is aware
+        vc = processed.get("versioncreated")
+        if not isinstance(vc, datetime):
+            processed["versioncreated"] = datetime.now(timezone.utc)
+        elif vc.tzinfo is None:
+            processed["versioncreated"] = vc.replace(tzinfo=timezone.utc)
+
+        # 4) Expiry based on provider config (hours)
```
petrjasek: that's handled automatically, pls remove it from parser

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        if isinstance(uri, (str, int)):
+            s = str(uri)
+            return f"urn:newsml:stt.fi:stt_tt_content_api:{hashlib.sha1(s.encode('utf-8')).hexdigest()}"
+        try:
+            blob = json.dumps(item, ensure_ascii=False, sort_keys=True)
+            h = hashlib.sha1(blob.encode("utf-8")).hexdigest()
+            return f"urn:newsml:stt.fi:stt_tt_content_api:{h}"
+        except Exception:
+            return f"urn:newsml:stt.fi:stt_tt_content_api:{uuid.uuid4()}"
```
petrjasek: not sure that's a good idea in general, that would create a new item for each version

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,140 @@
...
+    def _ensure_guid(self, item: Dict[str, Any]) -> str:
+        uri = (
+            item.get("uri")
+            or item.get("guid")
+            or item.get("original_id")
+            or item.get("_id")
+        )
+        if isinstance(uri, (str, int)):
+            s = str(uri)
+            return f"urn:newsml:stt.fi:stt_tt_content_api:{hashlib.sha1(s.encode('utf-8')).hexdigest()}"
```
petrjasek: I wonder why using hashlib and not using eg. `250815-usaryssmote6uv-06eb0f86` as suffix or converthing the whole url into uri?

minhpham212:
@petrjasek pls help me check issue 
image

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def can_parse(self, payload: Any) -> bool:
+        return isinstance(payload, dict) or (
+            isinstance(payload, list) and all(isinstance(i, dict) for i in payload)
+        )
+
+    def _ensure_guid(self, item: Dict[str, Any]) -> str:
+        """Generate GUIDs with a TT-specific namespace while respecting existing URNs."""
+        base_guid = super()._ensure_guid(item)
+        tt_prefix = "urn:newsml:stt.fi:stt_tt_content_api:"
```
petrjasek: imo the prefix should be like `urn:tt.fi:` it's not really STT content is it?

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,204 @@
...
+            "default": "200",
+            "description": "Safety cap for the number of pages to fetch when paginating. Range: 1-10000.",
+        },
+    ]
+
+    def _update(self, provider, update) -> Iterable[Dict]:
+        """
+        TT-specific update that fetches all pages and yields parsed dict items.
+        """
+        logger.info(
```
petrjasek: that's already logged when calling the service

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,204 @@
...
+        offset = 0
+        total = None
+
+        for page in range(max_pages):
+            # Merge/override pagination params each loop
+            qs = {**base_qs, "s": str(page_size), "fr": str(offset)}
+            page_url = urlunparse(parsed._replace(query=urlencode(qs, doseq=True)))
+
+            # Use base class HTTP retry infrastructure
+            response = self._get_with_retry(page_url, headers=headers, timeout=300)
```
petrjasek: timeout 300 is imo a bit too much, or is the api that slow?

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,204 @@
...
+            # Use base class JSON parsing with error handling
+            data = self._safe_json(response, provider)
+
+            if total is None and isinstance(data, dict):
+                # `total` may not always be present; handle gracefully
+                total = data.get("total")
+                logger.info("TT API: total from first page = %s", total)
+
+            batch = self._extract_tt_items_from_response(data)
+            logger.info(
```
petrjasek: that's rather for .debug

server/tests/stt_tt_content_api_test.py
```diff
@@ -0,0 +1,713 @@
...
+            # Should have processed all items
+            self.assertEqual(2, len(items))
+
+            # Check that items were parsed correctly
+            for parsed_item in items:
+                self.assertEqual("text", parsed_item["type"])
+                self.assertEqual("usable", parsed_item["pubstatus"])
+                self.assertIn("guid", parsed_item)
+
+    def test_fixture_data_structure(self):
```
petrjasek: why testing this?

server/tests/stt_tt_content_api_test.py
```diff
@@ -0,0 +1,713 @@
...
+        # Check original data is preserved
+        self.assertEqual(test_item["uri"], parsed_item["uri"])
+        self.assertEqual(test_item["headline"], parsed_item["headline"])
+
+        # Check GUID generation
+        self.assertTrue(
+            parsed_item["guid"].startswith("urn:newsml:stt.fi:stt_tt_content_api:")
+        )
+
+    def test_parser_content_expiry(self):
```
petrjasek: this is not needed imo

server/tests/stt_tt_parse_content_api_test.py
```diff
@@ -0,0 +1,138 @@
...
+        # Parser should return a list of dicts
+        self.assertIsInstance(result, list)
+        self.assertEqual(1, len(result))
+        parsed_item = result[0]
+
+        # Check required fields
+        self.assertEqual("text", parsed_item["type"])
+        self.assertEqual("usable", parsed_item["pubstatus"])
+        self.assertIn("guid", parsed_item)
+        self.assertIn("versioncreated", parsed_item)
```
petrjasek: pls test the actual values, not only that it's present, it could be wrong type or value

server/tests/stt_tt_parse_content_api_test.py
```diff
@@ -0,0 +1,138 @@
...
+        # Check original data is preserved
+        self.assertEqual(test_item["uri"], parsed_item["uri"])
+        self.assertEqual(test_item["headline"], parsed_item["headline"])
+
+        # Check GUID generation
+        self.assertTrue(
+            parsed_item["guid"].startswith("urn:newsml:stt.fi:stt_tt_content_api:")
+        )
+
+    def test_parser_content_expiry(self):
```
petrjasek: seems not needed as well

petrjasek (commented):
pls remove all those extra and duplicated tests, seems like lots of code is duplicated in those two test files, not sure if we need parser specific one if it checks the parsing in the other file. there is no value in testing something 2x, just more work to maintain it

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: I updated

server/tests/stt_tt_content_api_test.py
minhpham212: I removed tis test

server/tests/stt_tt_content_api_test.py
minhpham212: I remove this test

server/tests/stt_tt_parse_content_api_test.py
minhpham212: I updated

server/tests/stt_tt_parse_content_api_test.py
minhpham212: I remove this test

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: I change timeout to 60

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: I remove logs

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
minhpham212: I updated

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,203 @@
...
+                raise ParserError.parseMessageError(ex, provider, data=item)
+
+        # Final guard: ensure only dicts are returned
+        # (avoids filter_expired_items crash)
+        parsed_items = [it for it in parsed_items if isinstance(it, dict)]
+        return parsed_items
+
+    def _fetch_tt_data(self, provider) -> List[Dict]:
+        """
+        Fetch all items from TT Content API with pagination.
```
petrjasek, line 150: I wonder why actually fetching all items every time it runs, you could use the `trs` param with last updated timestamp to only get new content since last time

minhpham212: I updatet this

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -139,7 +165,48 @@ def _fetch_tt_data(self, provider) -> List[Dict]:
...
+            if isinstance(v, str):
+                s = v.strip().lower()
+                if s in {"1", "true", "yes", "y", "on"}:
+                    return True
+                if s in {"0", "false", "no", "n", "off", ""}:
+                    return False
+            return False
+
+        # Optional incremental sync using 'trs' (timestamp since last run)
+        use_trs = _coerce_bool(config.get("use_trs", False))
```
petrjasek: I wonder why making it optional, is it required not to be optional at some point?

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        base_guid = super()._ensure_guid(item)
+
+        target_prefix = "urn:newsml:stt.fi:stt_tt_content_api:"
+        if base_guid.startswith(target_prefix):
+            return base_guid
+
+        # Normalize TT and legacy STT GUID prefixes to the TT Content API namespace
+        prefix_mappings = [
+            ("urn:tt.fi:", target_prefix),
```
petrjasek: it's rather `tt.se` and not `tt.fi` my bad

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -0,0 +1,119 @@
...
+            ("urn:newsml:stt.fi:contentapi:", target_prefix),
+            ("urn:newsml:stt.fi:", target_prefix),
```
petrjasek: why are those here? would there be stt content ingested?

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: I updated this code

minhpham212:
@petrjasek  pls help me review

server/stt/io/feed_parsers/__init__.py
```diff
@@ -0,0 +1,3 @@
+import logging
+
+logging.basicConfig(level=logging.INFO)
```
petrjasek: you could do it once inside `io/__init__.py` and it will work for both parsers and services

server/tests/stt_tt_parse_content_api_test.py
petrjasek: where?

server/tests/stt_tt_parse_content_api_test.py
petrjasek: seems like it's still here

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
minhpham212: I removed this function

server/stt/io/feed_parsers/__init__.py
minhpham212: updated

minhpham212:
@petrjasek  pls check again

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
```diff
@@ -114,6 +89,23 @@ def _parse_one(self, src: Dict[str, Any], provider: dict) -> Dict[str, Any]:
             processed["body_html"] = body_html or ""
         return processed
 
+    def _ensure_guid(self, item: Dict[str, Any]) -> str:
+        """Ensure GUIDs use the TT-specific namespace when auto-generated."""
+        base_guid = super()._ensure_guid(item)
+        if not isinstance(base_guid, str):
```
petrjasek: what would it be if not a string?

server/tests/stt_tt_parse_content_api_test.py
petrjasek: can you change those 2 `assertIn` to `assertEqual` so it's visible what values are retured form the parser?

server/stt/io/feed_parsers/stt_tt_parse_content_api.py
minhpham212: I removed this function

minhpham212:
I updated this code , follow review, pls help me check

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,261 @@
...
+            "placeholder": "200",
+            "required": False,
+            "default": "200",
+            "description": (
+                "Safety cap for the number of pages to fetch when "
+                "paginating. Range: 1-10000."
+            ),
+        },
+        {
+            "id": "use_trs",
```
petrjasek: I still don't know why is this configurable?

server/tests/stt_tt_parse_content_api_test.py
```diff
@@ -0,0 +1,101 @@
...
+        parsed_item = result[0]
+
+        # Should have all required defaults
+        self.assertEqual("text", parsed_item["type"])
+        self.assertEqual("usable", parsed_item["pubstatus"])
+        self.assertIn("versioncreated", parsed_item)
+        self.assertEqual("Test minimal headline", parsed_item["headline"])
+        self.assertEqual("", parsed_item["body_html"])
+
+    def test_parser_guid_consistency(self):
```
petrjasek, line 44: imo one `assertEqual` test with `uri` would be enough

server/tests/stt_tt_parse_content_api_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # Parser should return lists
+        self.assertIsInstance(result1, list)
+        self.assertIsInstance(result2, list)
+        self.assertEqual(1, len(result1))
+        self.assertEqual(1, len(result2))
+
+        self.assertEqual(result1[0]["uri"], result2[0]["uri"])
+
+    def test_fixture_data_structure(self):
```
petrjasek: why testing the fixture?

server/tests/stt_parse_content_api_test.py
```diff
@@ -301,8 +301,8 @@ def test_parser_with_fixture_data(self):
 
         # Check specific values
         self.assertEqual(
-            "urn:newsml:stt.fi:contentapi:0c0ca4c14a785ce410944406f2aa55df9169d242ed620c38b06f194418c2934f",
-            parsed_item["guid"],
+            "https://stt-uat-api.superdesk.pro/contentapi/items/urn%3Anewsml%3Astt.fi%3A%3A107136785",
```
petrjasek, line 305: this should be rather `urn:newsml:stt.fi::107136785`

server/tests/stt_content_api_test.py
```diff
@@ -79,36 +75,13 @@ def test_all_fixture_items_parse_successfully(self):
                 self.assertIsInstance(parsed, dict)
                 self.assertEqual(parsed.get("type"), "text")
                 self.assertEqual(parsed.get("pubstatus"), "usable")
-                self.assertIsInstance(parsed.get("guid"), str)
-                self.assertTrue(parsed["guid"])  # non-empty
+                self.assertIsInstance(parsed.get("uri"), str)
+                self.assertTrue(parsed["uri"])  # non-empty
```
petrjasek: one `assertEqual` would be better than checking instance and non empty

server/tests/stt_parse_content_api_test.py
minhpham212: I am using the original URI so that it does not change the data

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: Users can enable or disable this configuration.

minhpham212:
@petrjasek, please check this configuration library issue
The conflict is caused by:
    The user requested tzlocal==2.1
    superdesk-core 3.2.0.dev0 depends on tzlocal>=5.

petrjasek:
I've merged the async version in so the tests should be working now

server/stt/io/feeding_services/stt_tt_content_api.py
petrjasek: but why? I don't see it as a requirement and it complicates things

minhpham212:
@petrjasek pls help me check again.

petrjasek:
I checked it, would still remove that `use_trs` config and use it always

minhpham212:
@petrjasek I remove use_trs

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,253 @@
...
+    ]
+
+    async def _update(self, provider, update) -> Iterable[Dict]:
+        """
+        TT-specific update that fetches all pages and yields parsed dict items.
+        Async to match the base class contract. Internally, the HTTP fetching is
+        still synchronous, so we offload it via asyncio.to_thread.
+        """
+        # Offload sync fetch to a worker thread to avoid blocking the event loop
+        json_items = await asyncio.to_thread(self._fetch_tt_data, provider, update)
```
petrjasek: not sure if that's useful here, would be probably better to just use [https://docs.aiohttp.org/en/stable/](https://docs.aiohttp.org/en/stable/) for fetching, the rest can run as usual

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -0,0 +1,253 @@
...
+        items: List[Dict] = []
+        offset = 0
+        total = None
+
+        for page in range(max_pages):
+            # Merge/override pagination params each loop
+            qs = {**base_qs, "s": str(page_size), "fr": str(offset)}
+            if trs_value:
+                qs["trs"] = trs_value
+            page_url = urlunparse(parsed._replace(query=urlencode(qs, doseq=True)))
```
petrjasek: I wonder why doing this urlencode urlunparse?

minhpham212: updated

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: updated

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: @petrjasek  Please help me check the CI/CD issue.

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -249,5 +250,94 @@ def _extract_tt_items_from_response(self, data) -> List:
...
+                            if total is None and isinstance(data, dict):
+                                total = data.get("total")
+                            batch = self._extract_tt_items_from_response(data)
+                            if isinstance(batch, list) and batch:
+                                items.extend(
+                                    [it for it in batch if isinstance(it, dict)]
+                                )
+                            else:
+                                # No more results
+                                return [it for it in items if isinstance(it, dict)]
```
petrjasek: not sure if it needs another iteration if you only add `dicts` to `items`

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -249,5 +250,94 @@ def _extract_tt_items_from_response(self, data) -> List:
...
+                            break
+                    except Exception as ex:
```
petrjasek: this should only rerun on that `aiohttp.ClientResponseError` right?

server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -249,5 +250,94 @@ def _extract_tt_items_from_response(self, data) -> List:
...
+                        attempt += 1
+                        if attempt >= 3:
+                            raise ParserError.parseMessageError(ex, provider, data=None)
+                        await asyncio.sleep(2 ** (attempt - 1))
+                offset += page_size
+                if isinstance(total, int) and offset >= total:
+                    break
+        return [it for it in items if isinstance(it, dict)]
```
petrjasek, line 356: same issue like on line 330

minhpham212: updated

server/stt/io/feeding_services/stt_tt_content_api.py
minhpham212: updated

petrjasek (approved):

PR #354, opened 2025-08-17
implement task STT-95 TT NINJS parser

minhpham212:
### Summary

Add a TT ninjs parser and TT→STT metadata mapping (parity with our NewsML/XSLT import), plus tests. Standardizes HTML cleanup and datetime (Europe/Helsinki → UTC). Lint/format green.

### Changes
	•	Parser: server/stt/stt_tt_ninjs.py
	•	Class: STTTTNINJSFeedParser (registry: stt_tt_ninjs)
	•	Transform rules:
	•	Prefer body_html5, fallback body_html/body
	•	Sanitize HTML: <div class="byline">→<p class="byline">, <span>→<p>, strip wrappers (html, body, article, section, …); return inner HTML
	•	Datetime: naive → Europe/Helsinki → UTC ISO; tz-aware → UTC ISO
	•	Mapping:
	•	source = "TT", desk = "Ulkomaat"
	•	Department from department|dept|sector|profile → STT id/name (e.g., UTR→14/Ulkomaat, SPT→16/Urheilu, default 3/Kotimaa)
	•	priority from urgency (int)
	•	name from headline|title
	•	external_id from originaltransmissionreference (fallback id|guid|uri|job)
	•	description_text from description_text|filename

Tests
	•	File: server/tests/stt_tt_new_ninjs_test.py
	•	Fixture: server/tests/fixtures/json/stt_new_ninjs.json
	•	Covers: can_parse (reject type="image"), mapping


server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,238 @@
...
+            or ninjs_local.get("sector")  # sometimes codes come here
+            or ninjs_local.get("profile")  # last-resort guess
+        )
+        dept_id, dept_name = self._map_department(tt_dept_code)
+        # Store both machine and human-friendly values. Adapt field names to your schema.
+        item.setdefault("extra", {})
+        item["extra"].setdefault("stt_meta", {})
+        item["extra"]["stt_meta"]["department_id"] = dept_id
+        item["extra"]["stt_meta"]["department_name"] = dept_name
+        item["extra"]["stt_meta"]["tt_department_code"] = tt_dept_code
```
petrjasek, line 216: afaik we want to store stt department as `anpa_category`, it should probably find the matching value from CV and get the name from there

minhpham212: I updated the item to use anpa_category.

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,137 @@
...
+            self.assertEqual(item["extra"]["stt_meta"]["department_id"], 14)
+            self.assertEqual(item["extra"]["stt_meta"]["department_name"], "Ulkomaat")
+            self.assertEqual(item["extra"]["stt_meta"]["tt_department_code"], "UTR")
+            # anpa_category should be populated from CV mapping (fallback name to map)
+            self.assertIn("anpa_category", item)
+            self.assertEqual(
+                item["anpa_category"], [{"qcode": "ulkomaat", "name": "Ulkomaat"}]
+            )
+            self.assertEqual(item["priority"], 5)
+            self.assertEqual(item["name"], "My Headline")
```
petrjasek: this should go to `item["headline"]` probably

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,137 @@
...
+            self.assertEqual(item["external_id"], "ext-123")
+            self.assertEqual(item["description_text"], "file.txt")
+            self.assertNotIn("associations", item)
+            self.assertNotIn("<html", item["body_html"])
+            self.assertNotIn("<body", item["body_html"])
+            self.assertIn('<p class="byline">Byline</p>', item["body_html"])
+            self.assertIn("<p>Intro</p>", item["body_html"])
```
petrjasek: would be good to also test that `type` and `versioncreated` are set properly, those are important in the ingest process

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -1,137 +1,59 @@
...
-
-        sanitized = parser.sanitise_stt_tt_html(html)
-
-        self.assertEqual(sanitized, "")
+        fixture = os.path.join(dirname, "fixtures", self.fixture)
+        provider = {"name": "Test"}
+        with self.ctx:
+            with open(fixture, "r", encoding="utf-8") as f:
+                parser = self.parser_class()
+                self.json_data = json.load(f)
```
petrjasek: I guess this is the only difference with base class, might be better to introduce a method in base class like load_file() and use that with the parser, so it can override only that part without duplicating the whole method

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -1,137 +1,59 @@
...
+        self.assertIn("desk", self.item)
+        self.assertIn("extra", self.item)
+        self.assertIn("stt_meta", self.item["extra"])
+        self.assertEqual(self.item["source"], "TT")
+        self.assertEqual(self.item["desk"], "Ulkomaat")
+        self.assertEqual(self.item["extra"]["stt_meta"]["department_id"], 3)
+        self.assertEqual(self.item["extra"]["stt_meta"]["department_name"], "Kotimaa")
+        self.assertIsNone(self.item["extra"]["stt_meta"]["tt_department_code"])
+        self.assertEqual(
+            self.item["anpa_category"], [{"qcode": "kotimaa", "name": "Kotimaa"}]
```
petrjasek: `qcode` should be department id so `3`

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -1,137 +1,59 @@
...
+
+    def test_item_type_and_version_fields(self):
+        self.assertEqual(self.item["type"], "picture")
+        self.assertIn("guid", self.item)
+        self.assertIn("uri", self.item)
+        self.assertIn("mimetype", self.item)
+
+    def test_external_id_and_description(self):
+        self.assertEqual(
+            self.item["external_id"],
```
petrjasek: we don't have such a field in schema, but you can store it as `uri`

server/tests/stt_tt_new_ninjs_test.py
minhpham212: updated

server/tests/stt_tt_new_ninjs_test.py
minhpham212: I updated this code

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,69 @@
...
+            self.item["anpa_category"], [{"qcode": "3", "name": "Kotimaa"}]
+        )
+
+    def test_body_html_and_byline(self):
+        html = self.item.get("body_html", "")
+        self.assertEqual(html, "")
+        self.assertIn("byline", self.item)
+        self.assertEqual(self.item["byline"], "Fredrik Sandberg/TT")
+        self.assertEqual(self.item["type"], "picture")
+        # versioncreated is not set by the parser for picture items
```
petrjasek: would probably rather set it based on the text item, it can't be later than that, otherwise if using current time it might show up always on top in lists

minhpham212:
@petrjasek  pls help me review this branch

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,428 @@
...
+
+        if current_dt and parent_dt and current_dt > parent_dt:
+            item["versioncreated"] = parent_dt
+
+    def _url_ok(self, url: str) -> bool:
+        """Return True if URL is publicly accessible (HTTP 200). Uses HEAD then GET fallback."""
+        if not url or not isinstance(url, str) or not url.startswith("http"):
+            return False
+        try:
+            resp = requests.head(
```
petrjasek: would be good to use session here

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def _transform_from_ninjs(self, ninjs: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Extend base NINJS transformation with TT/STT specifics:
+        - Detect sports sector
+        - Sanitize/normalize HTML (prefer body_html5)
+        - Map TT metadata -> STT metadata (Source, Department, Desk, Priority, Name, ExternalID)
+        - Drop heavy associations after parent mapping
+        """
+        self.is_sport_item = ninjs.get("sector") == "SPT"
```
petrjasek: what's the usecase for this?

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,428 @@
...
+        item["body_html"] = self.sanitise_stt_tt_html(raw_html)
+
+        # --- Metadata mapping TT -> STT (derivative of Neo import rules)
+
+        # Source: fixed "TT"
+        item["source"] = "TT"
+
+        # Desk: fixed "Ulkomaat" (per spec)
+        # If your schema uses another field name (e.g. "task.desk"), adjust here.
+        item["desk"] = "Ulkomaat"
```
petrjasek: we don't have such field, if it should go to specific desk that's configured via routing rules, so not sure if that's needed at all?

server/stt/stt_tt_new_parse_ninjs.py
petrjasek: what's the usecase for that `stt_meta` data?

server/tests/stt_tt_new_ninjs_test.py
petrjasek: how is this resolved now?

server/stt/stt_tt_new_parse_ninjs.py
minhpham212: I noticed this field exists in stt_tt_ninjs.py. Is it necessary?

server/stt/stt_tt_new_parse_ninjs.py
minhpham212: I remove this code

server/stt/stt_tt_new_parse_ninjs.py
minhpham212: updated

server/tests/stt_tt_new_ninjs_test.py
minhpham212: I updated more test case for "versioncreated"

server/stt/stt_tt_new_parse_ninjs.py
petrjasek: looks like it's not used anywhere

server/stt/stt_tt_new_parse_ninjs.py
minhpham212: Should we be storing this additional data?

server/stt/stt_tt_new_parse_ninjs.py
minhpham212: I removed this code

server/stt/stt_tt_new_parse_ninjs.py
petrjasek: I don't see why we would need to

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,155 @@
...
+        parser = self.parser_class()
+        result = parser._transform_from_ninjs(test_data)
+
+        # The versioncreated should remain unchanged since no parent text item exists
+        from dateutil.parser import isoparse
+
+        expected_time = isoparse("2024-10-03T16:00:00+02:00")
+        self.assertEqual(result["versioncreated"], expected_time)
+
+    def test_versioncreated_capping_with_multiple_parent_candidates(self):
```
petrjasek: I don't think a picture can be a parent to multiple text items, the link is usually the other way around.
or is this a real scenario? is there such content?

minhpham212: I am using real data from json file to make a test.

minhpham212:
@petrjasek pls help me review

server/tests/stt_tt_new_ninjs_test.py
petrjasek: in the json file there is text item with associated picture, not the other way around. or are you referring to some other file?

minhpham212: https://github.com/superdesk/superdesk-stt/pull/354/files#diff-ee077a5b339c79494c3e37bb798f7f991542c619e58a84cf848e150ca4ccd136

minhpham212: @petrjasek pls help me check this code

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertEqual(html, "")
+        self.assertIn("byline", self.item)
+        self.assertEqual(self.item["byline"], "Fredrik Sandberg/TT")
+        self.assertEqual(self.item["type"], "picture")
+
+    def test_item_type_and_version_fields(self):
+        self.assertEqual(self.item["type"], "picture")
+        self.assertIn("guid", self.item)
+        self.assertIn("uri", self.item)
+        self.assertIn("mimetype", self.item)
```
petrjasek: would be good to test the actual values, being present doesn't show you it actually works, item like:
```
{"type": "picture", "guid": null, "uri": null, "mimetype": null}
```
would pass the test, but that wouldn't be working parser

server/settings.py
```diff
@@ -143,7 +143,7 @@ def env(variable, fallback_value=None):
     "stt.macros",
     "stt.search_providers.newshub",
     "stt.ai_proxy",
-    "stt.paivalista_filters",
```
petrjasek: this line should stay probably

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,441 @@
...
+from superdesk.io.registry import register_feed_parser
+from superdesk.text_utils import sanitize_html
+
+TIMEZONE = "Europe/Helsinki"
+
+IGNORE_REMOTE_IMAGES = False  # Only ignore when public URL is not accessible
+IMAGE_CHECK_TIMEOUT = 2  # seconds
+
+# Controlled Vocabulary id used in Superdesk for STT departments
+STT_DEPT_VOCAB_ID = "stt_department_categories"
```
petrjasek: it's using different ids in different parsers, should be probably consistent, maybe in some shared file

minhpham212: I made a constants.py file for this refactor.

server/tests/stt_tt_new_ninjs_test.py
minhpham212: I updated test to assertEqual

minhpham212:
@petrjasek pls help me approve :D

minhpham212:
@petrjasek  pls check help me

server/stt/constants.py
```diff
@@ -0,0 +1,27 @@
+"""Shared constants for STT parsers to ensure consistency across all feed parsers."""
+
+# Vocabulary IDs used in Superdesk for STT content
+STT_DEPARTMENT_VOCAB_ID = "sttdepartment"
+STT_MEDIA_TOPICS_VOCAB_ID = "stt_media_topics"
```
petrjasek, line 5: I don't see it being used here, but it should be `topics` (matching the actual CV id)

server/tests/stt_tt_new_ninjs_test.py
```diff
@@ -0,0 +1,162 @@
...
+    def test_body_html_and_byline(self):
+        html = self.item.get("body_html", "")
+        # Picture items don't have body_html content
+        self.assertEqual(html, "")
+        self.assertIn("byline", self.item)
+        self.assertEqual(self.item["byline"], "Fredrik Sandberg/TT")
+        self.assertEqual(self.item["type"], "picture")
+
+    def test_item_type_and_version_fields(self):
+        self.assertEqual(self.item["type"], "picture")
```
petrjasek: imo this one should be `text` looking at the sample, and `assoc["type"]` should be `picture`, or am I missing something?

minhpham212: I changed the image to text.

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,410 @@
...
+            item["description_text"] = description
+
+        if IGNORE_REMOTE_IMAGES:
+            item.pop("associations", None)
+
+        return item
+
+    # --------- Helpers ---------
+
+    def _get_cv_item_by_qcode(
```
petrjasek: I don't see it being used

server/stt/stt_tt_new_parse_ninjs.py
```diff
@@ -0,0 +1,410 @@
...
+            return None
+        try:
+            dt = isoparse(value)
+            if dt.tzinfo is None:
+                return dt.replace(tzinfo=tz.gettz(STT_TIMEZONE)).astimezone(tz.UTC)
+            return dt.astimezone(tz.UTC)
+        except Exception:
+            return None
+
+    def _cap_versioncreated_to_parent(
```
petrjasek: I don't see this being used

petrjasek (approved):

PR #362, opened 2025-08-27
update stt parsers metadata handling

petrjasek:
- map stt topics to media topics
- map location to sttplace
- map department to anpa_category
- parse urgency
- parse source as sttsource
- set language en for translated content

STT-1268


eos87 (approved):
Looks good

PR #367, opened 2025-08-31
Feature/st 84 info porssi

minhpham212:
•	New parser class STTInfoPorssi:
•	Extracts and maps fields from the provided XML feed.
•	Generates guid in the format stt:announcement:<id>-<language>.
•	Ensures language detection (defaults to "fi" for Finnish).
•	Sets source as "STT" and priority as 3.
•	Assigns default body_html placeholder: <p>Handled by XSLT</p>.
•	Metadata handling:
•	Populates required fields: guid, name, source, priority, language, original_xml.
•	Fills description from name.
•	Adds task object with desk = "Kotimaa".
•	Includes anpa_category and subject fields.
•	Leaves slugline and headline empty by design.
•	Unit tests:
•	Added stt_info_porssi_test.py verifying:
•	Correct headline extraction from fixture.
•	Proper GUID generation.
•	Language detection.
•	Required fields presence.
•	Placeholder body_html.
•	Additional metadata integrity.


minhpham212:
I changed name or brach correct with task STT-84

minhpham212:

server/stt/stt_info_porssi.py
```diff
@@ -0,0 +1,155 @@
...
+            selected = None
+            for news_item in news_items:
+                try:
+                    lang = news_item.xpath(
+                        "NewsComponent/DescriptiveMetadata/Language/@FormalName"
+                    )[0]
+                except IndexError:
+                    logger.warning(
+                        "missing language in item, ignoring it.\nxml: {xml}".format(
+                            xml=etree.tostring(news_item, encoding="unicode")
```
petrjasek: would rather add some id only and not dump the whole xml in logs

server/stt/stt_info_porssi.py
```diff
@@ -0,0 +1,155 @@
...
+                "NewsComponent/AdministrativeMetadata/Source/Party/@FormalName"
+            )
+            source = source_elements[0] if source_elements else "STT"
+
+            headline = selected.findtext("NewsComponent/NewsLines/HeadLine") or ""
+
+            item = {
+                "guid": f"stt-info-porssi_{guid}",
+                "headline": headline,
+                "slugline": headline,
```
petrjasek, line 139: is that typo or by design? if that's by design pls add a comment

server/stt/stt_info_porssi.py
```diff
@@ -0,0 +1,155 @@
...
+                "body_html": body,
+                "source": source,
+                "priority": 3,
+                "language": lang,
+                "anpa_category": [],
+                "subject": [],
+                "type": "text",
+                "versioncreated": datetime.utcnow().isoformat() + "Z",
```
petrjasek: there is a timestamp in `NewsManagement.ThisRevisionCreated`

server/stt/stt_info_porssi.py
```diff
@@ -0,0 +1,155 @@
...
+            # Use xpath() for attribute selectors
+            source_elements = selected.xpath(
+                "NewsComponent/AdministrativeMetadata/Source/Party/@FormalName"
+            )
+            source = source_elements[0] if source_elements else "STT"
+
+            headline = selected.findtext("NewsComponent/NewsLines/HeadLine") or ""
+
+            item = {
+                "guid": f"stt-info-porssi_{guid}",
```
petrjasek: `urn:` might be a better prefix for it, but you could also use that `PublicIdentifier` value as is

server/tests/stt_info_porssi_test.py
```diff
@@ -0,0 +1,56 @@
...
+        self.assertEqual(self.item["priority"], 3)
+
+    def test_language_detected(self):
+        """Test that language is detected correctly as Finnish."""
+        self.assertEqual(self.item["language"], "fi")
+
+    def test_type_and_versioncreated(self):
+        """Test that type and versioncreated are set correctly."""
+        self.assertEqual(self.item["type"], "text")
+        self.assertIsInstance(self.item["versioncreated"], str)
```
petrjasek: it should be `datetime` instance

server/stt/stt_info_porssi.py
minhpham212: updated

server/stt/stt_info_porssi.py
minhpham212: updated

server/tests/stt_info_porssi_test.py
minhpham212: updated

server/stt/stt_info_porssi.py
minhpham212: updated

server/stt/stt_info_porssi.py
```diff
@@ -0,0 +1,213 @@
...
+            public_id = selected.findtext(
+                "Identification/NewsIdentifier/PublicIdentifier"
+            )
+            news_item_id = selected.findtext("Identification/NewsIdentifier/NewsItemId")
+            if public_id:
+                guid_value = f"urn:stt:info-porssi:{public_id}"
+            elif news_item_id:
+                guid_value = f"urn:stt:info-porssi:{news_item_id}"
+            else:
+                guid_value = (
```
petrjasek: seems like it's checking before that there is an id so this won't be ever used if I got it right, would be good to remove it then

minhpham212: I update this code

petrjasek (approved):

PR #384, opened 2025-09-08
Fix vocabulary config in assignments list

tomaskikutis:
STT-1242

I've fixed the config format, but vocabulary fields still won't work, because `subject` field doesn't come from `coverage` to `assignment` (`assignment.planning.subject`). Could you fix this @petrjasek?

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -12638,8 +12638,7 @@
         },
         "node_modules/superdesk-planning": {
             "version": "3.0.0-dev.0",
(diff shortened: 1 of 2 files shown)


petrjasek (approved):

petrjasek:
looking at the subject and seems like the issue is it's stored as `coverage.subject` and not `coverage.planning.subject` like the other fields, can you fix it @thecalcc ? based on https://github.com/superdesk/superdesk-planning/pull/2296

thecalcc:
@tomaskikutis now it should be good. I've updated with latest planning that has subject in `coverage.planning`

PR #387, opened 2025-09-10
[STT-1318] - Handle the priority / stturgency mapping in the output to NH

BrianMwangi21:
# Purpose
Handle the STT urgency mapping in `update_ninjs_stt_urgency` for planning items output

# What has changed
- Added `update_ninjs_stt_urgency` which maps priority codes to the stturgency names
- Handles special case where priority 2 and 3 are mapped to stturgency-2

### DIFF
--- a/server/stt/stt_json_planning_formatter.py
+++ b/server/stt/stt_json_planning_formatter.py
@@ -1,9 +1,5 @@
-import logging
-
 from superdesk import get_resource_service
 from planning.output_formatters import JsonPlanningFormatter
-
-logger = logging.getLogger(__name__)
 
 
 class STTJsonPlanningFormatter(JsonPlanningFormatter):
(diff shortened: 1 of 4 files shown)


petrjasek:
this actually needs a new formatter because it should work for planning items, we have some core JSON Planning formatter in superdesk-planning that needs to be extended

server/stt/stt_json_planning_formatter.py
```diff
@@ -0,0 +1,58 @@
...
+
+        stturgency_item = None
+        for vocab_item in vocabulary_items:
+            if vocab_item.get("qcode") == qcode:
+                stturgency_item = vocab_item
+                break
+
+        if stturgency_item:
+            if "subject" not in item:
+                item["subject"] = []
```
petrjasek: you can do `item.setdefault("subject", [])`

server/stt/stt_json_planning_formatter.py
```diff
@@ -0,0 +1,58 @@
+import logging
+
+from superdesk import get_resource_service
+from planning.output_formatters import JsonPlanningFormatter
+
+logger = logging.getLogger(__name__)
```
petrjasek: better remove it if it's not used

petrjasek (approved):

PR #410, opened 2025-09-18
feat(ingest): add STT since-based HTTP feeding service using NinJS parser (mainly for Sport results / stats)

0is1:
* Wire new feeding service app into INSTALLED_APPS
* Implement STTWithSinceHTTPFeedingService (NAME: "stt_since_http") for STT Sport Results API
* Build URL with since=<ISO-8601 Z> from provider.last_updated (fallback to now)
* Auth via Bearer token (auth_token in provider.config), requests.Session with timeout
* Parse response by feeding one NinJS item at a time; supports dict, {items: [...]}, or list payloads
* Normalize timestamps (versioncreated/firstcreated/embargoed) to UTC "Z" to satisfy ninjs datetime parsing
* Adapt categories: map anpa_category → ninjs "service" and normalize service to [{code, name?, scheme?}]

### DIFF
--- a/server/stt/io/feeding_services/stt_http_with_since.py
+++ b/server/stt/io/feeding_services/stt_http_with_since.py
@@ -182,16 +182,20 @@ class STTWithSinceHTTPFeedingService(HTTPFeedingService):
             data = json.loads(content)
         except Exception:
             # Fallback: let parser.parse handle the raw file content by writing it to disk
-            with NamedTemporaryFile("wb", delete=False, suffix=".json") as f:
+            # Since this runs inside Docker (Linux/POSIX), we can parse while the temp file is open
+            # and rely on automatic deletion when the context exits.
+            with NamedTemporaryFile("wb", delete=True, suffix=".json") as f:
+                logger.warning(
+                    "Falling back to raw file parsing due to JSON decode error"
+                )
                 f.write(content)
-                tmp = f.name
-            try:
-                results = self._call_parser(parser, tmp, provider)
-            finally:
-                try:
-                    os.unlink(tmp)
-                except Exception:
-                    pass
+                # Make sure the parser, which opens f.name immediately, sees the full contents:
+                # - flush(): push Python's buffer to the OS so size/data are visible to a new open()
+                # - fsync(): ask the OS to flush page cache + metadata; helpful on some FS/overlay layers
+                #   (Docker/overlayfs) to avoid stale size/content. Safe but optional for POSIX; drop if perf-critical.
+                f.flush()
+                os.fsync(f.fileno())
+                results = self._call_parser(parser, f.name, provider)
             return results
 
         # Normalize to a list of item dicts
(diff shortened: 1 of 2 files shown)


server/stt/io/feeding_services/stt_http_with_since.py
```diff
@@ -0,0 +1,358 @@
...
+        except Exception:
+            # Fallback: let parser.parse handle the raw file content by writing it to disk
+            with NamedTemporaryFile("wb", delete=False, suffix=".json") as f:
+                f.write(content)
+                tmp = f.name
+            try:
+                results = self._call_parser(parser, tmp, provider)
+            finally:
+                try:
+                    os.unlink(tmp)
```
petrjasek: imo if you move that call to `_call_parser` inside `with NamedTemporaryFile` block you can keep that default delete behaviour and you don't need to handle the cleanup after

server/stt/io/feeding_services/stt_http_with_since.py
```diff
@@ -0,0 +1,358 @@
...
+                        if el.get("name"):
+                            entry["name"] = el.get("name")
+                        if el.get("scheme"):
+                            entry["scheme"] = el.get("scheme")
+                        normalized.append(entry)
+                elif isinstance(el, str):
+                    normalized.append({"code": el})
+            out["service"] = normalized
+        elif isinstance(out.get("service"), str):
+            out["service"] = [{"code": out.get("service")}]
```
petrjasek: I wonder what's the use for the `service` value? we don't use it in superdesk there we have `anpa_category`

0is1: Yeah, actually that is completely unnecessary logic for all the "else"-blocks. I removed them in the latest commit. 

But I just noticed that the default ninjs parser expects `service` key that is then mapped to the `anpa_category` (https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/feed_parsers/ninjs.py#L101-L102) so that's why I added this kind of helper here. Do you have any idea why it's like that?

petrjasek: my bad, didn't realize it will call that ninjs parser on that, makes sense then

server/stt/io/feeding_services/stt_http_with_since.py
```diff
@@ -0,0 +1,331 @@
...
+            with NamedTemporaryFile("wb", delete=True, suffix=".json") as f:
+                logger.warning(
+                    "Falling back to raw file parsing due to JSON decode error"
+                )
+                f.write(content)
+                # Make sure the parser, which opens f.name immediately, sees the full contents:
+                # - flush(): push Python's buffer to the OS so size/data are visible to a new open()
+                # - fsync(): ask the OS to flush page cache + metadata; helpful on some FS/overlay layers
+                #   (Docker/overlayfs) to avoid stale size/content. Safe but optional for POSIX; drop if perf-critical.
+                f.flush()
```
petrjasek, line 196: I think this might require rather `f.seek(0)` after write so that parser car read the file from the beginning

petrjasek (approved):

PR #446, opened 2025-10-07
merging develop to next and converting parsers to async

petrjasek:


petrjasek:
hi @MarkLark86 can you just check the changes in https://github.com/superdesk/superdesk-stt/pull/446/commits/f6078a608aa4187a849b75ed0f0611b94548c911 ?

server/stt/io/feeding_services/stt_content_api.py
```diff
@@ -0,0 +1,274 @@
+from __future__ import annotations
+
+
+import logging
+from datetime import datetime
+from typing import Any, Dict, Iterable, List, Optional, Tuple
+import requests
```
MarkLark86, line 7: Ideally this would use aiohttp, so this request is not holding up the ASGI request queue.
Would this be something we do now or at a later date?

server/stt/lupaus_export/enrich_related.py
```diff
@@ -0,0 +1,298 @@
...
+        logger.exception(
+            "Error using service for %s (%s: %s).",
+            resource,
+            e.__class__.__name__,
+            e,
+        )
+        svc = None
+
+    # 1) service.get_from_mongo with (req=None, ...)
+    if svc and hasattr(svc, "get_from_mongo"):
```
MarkLark86, line 114: This function uses `get_from_mongo` and `get` methods of the service.
Ideally we should use the async variants instead, i.e. `get_from_mongo_async` and `get_async`

server/stt/stt_parse_lottery_veikkaus.py
```diff
@@ -0,0 +1,169 @@
...
+        self, file_path: str, provider: Optional[dict] = None
+    ) -> List[Dict[str, Any]]:
+        """Parse the Veikkaus file into a Superdesk ingest item.
+
+        Returns a list with a single item.
+        """
+        # Provider is part of the base signature; not used here.
+        del provider
+
+        print("PARSING", file_path)
```
MarkLark86: I'm guessing there were for debugging!

MarkLark86 (commented):
Looks good, just a couple small improvements (not sure if we do them now or later)

server/stt/io/feeding_services/stt_content_api.py
petrjasek: this is running in celery, so wouldn't change it much now, they might still change it anyhow

PR #477, opened 2025-10-21
Feature/oikaisut korjaukset export

0is1:
Add Oikaisut/Korjaukset planning export with helpers and filters
    
    - register oikaisut_korjaukset template in settings/data
    - add enrichment helpers to load related published corrections and skip drafts
    - expand shared template filters with HTML helpers and "6.25" time format
    - create Jinja template rendering grouped corrections and oikaisut output

    Tweak Lupaus export
    
    - Use exclude_drafts to remove any items in draft state
    - Add '6: "Vain tsekkaus"' to get_priority_from_agenda_item
    - Clean duplicate


server/stt/helpers/template_helpers.py
```diff
@@ -0,0 +1,17 @@
+"""Shared template helper functions for STT."""
+
+from __future__ import annotations
```
petrjasek: seems like this is [deprecated](https://docs.python.org/3/library/__future__.html#id1)

petrjasek (approved):

PR #493, opened 2025-10-28
[STT-1422] - Adjust JSON output format for Newshub

BrianMwangi21:
### Purpose
This PR adjusts the STT JSON output with the expectations by removing internal fields from planning and coverage fields and mapping agendas from anpa_category

### What has changed
- Added function to remove internal fields from the planning item.
- Added function to remove internal fields from the coverage items.
- Added function to map the anpa_category into the agendas field.
 
 ### 
 [STT-1422](https://sofab.atlassian.net/browse/STT-1422)

[STT-1422]: https://sofab.atlassian.net/browse/STT-1422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/stt/stt_json_planning_formatter.py
+++ b/server/stt/stt_json_planning_formatter.py
@@ -6,6 +6,44 @@ class STTJsonPlanningFormatter(JsonPlanningFormatter):
     name = "STT JSON Planning"
     type = "stt_json_planning"
 
+    CATEGORY_QCODE_NHUB_MAP = {
+        "3": {
+            "name": "Kotimaa",
+            "agenda_id": "673f47881562ef05527e195a",
+            "calendar_qcode": "kotimaa",
+        },
+        "4": {
+            "name": "Kulttuuri",
+            "agenda_id": "673f478f2c95fe4cab8ee67e4",
+            "calendar_qcode": "kulttuuri",
+        },
+        "9": {
+            "name": "Politiikka",
+            "agenda_id": "673f47a1f16a90a38481291f",
+            "calendar_qcode": "politiikka",
+        },
+        "11": {
+            "name": "Talous",
+            "agenda_id": "673f47a914ae7505c4ba6138",
+            "calendar_qcode": "talous",
+        },
+        "14": {
+            "name": "Ulkomaat",
+            "agenda_id": "673f47b114ae7505c4ba613a",
+            "calendar_qcode": "ulkomaat",
+        },
+        "16": {
+            "name": "Urheilu",
+            "agenda_id": "673f47ba1562ef05527e195c",
+            "calendar_qcode": "urheilu",
+        },
+        "not_sports": {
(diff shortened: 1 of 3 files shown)


server/stt/stt_json_planning_formatter.py
```diff
@@ -44,8 +44,69 @@ def update_stt_urgency(self, item):
...
+                    for field in exclude_fields:
+                        if field in planning:
+                            del planning[field]
+                    # Then filter out internal fields from the fields array
+                    if "fields" in planning:
+                        planning["fields"] = [
+                            f
+                            for f in planning["fields"]
+                            if f.get("field") not in exclude_fields
+                        ]
```
MarkLark86: To improve readability of this function, and to avoid nested ifs, we can use the return early pattern.

```suggestion
    def exclude_internal_coverage_fields(self, item):
        if not item.get("coverages"):
            return

        exclude_fields = [
            "headline",
            "location",
            "sttpicturewhatabout",
            "sttpicturewhatisphotographed",
            "internal_note",
            "sttdoesphotographerknow",
            "sttpictureregistrationok",
            "sttregistrationinfo",
            "sttpicturecategory",
            "sttpictureheadlinefi",
            "sttpicturecaptionfi",
            "sttpictureinstructionsfi",
            "sttpicturekeywordsfi",
            "sttpictureheadlineen",
            "sttpicturecaptionen",
            "sttpictureinstructionsen",
            "sttpicturekeywordsen",
            "sttpictureinvoiced",
        ]

        for coverage in item["coverages"]:
            planning = coverage.get("planning")
            if planning:
                continue

            # First remove direct attributes in the planning obj
            for field in exclude_fields:
                if field in planning:
                    del planning[field]

            # Then filter out internal fields from the fields array
            if "fields" in planning:
                planning["fields"] = [
                    f
                    for f in planning["fields"]
                    if f.get("field") not in exclude_fields
                ]
```

server/stt/stt_json_planning_formatter.py
```diff
@@ -44,8 +44,69 @@ def update_stt_urgency(self, item):
...
+
+    def map_agenda_output(self, item):
+        anpa_category = item.get("anpa_category", [])
+        agendas = []
+        for category in anpa_category:
+            agendas.append({"_id": category["name"], "name": category["name"]})
```
MarkLark86: I know the ticket suggests to use `name` for the `_id`. In CVs, `name` can change but we don't support changing the `qcode`. Therefor it might be good to use `qcode` here for `_id` instead (I've placed a comment in the ticket)

MarkLark86 (commented):
Looks good. Just 1 improvement suggestion, and then the question in linked ticket regarding `qcode` or `name` from CVs.

MarkLark86 (approved):

PR #497, opened 2025-10-30
[STT-1415] - Adjust STT Newsroom NINJS output format for Newshub

BrianMwangi21:
### Purpose
This PR adjusts the STT Newsroom NINJS output for newshub

### What has changed
- Added function to map place values to locators
- Added function to update body from viiva and sms content profiles
- Added function to update subheadline for pikaplus content profile
- Added function to update sttversion
- Added function to append sttnewsroomnote to ednote
 
###  [STT-1415](https://sofab.atlassian.net/browse/STT-1415)

[STT-1415]: https://sofab.atlassian.net/browse/STT-1415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/stt/stt_newsroom_ninjs_formatter.py
+++ b/server/stt/stt_newsroom_ninjs_formatter.py
@@ -1,4 +1,5 @@
 import logging
+import re
 
 from superdesk.publish.formatters import NewsroomNinjsFormatter
 
@@ -14,7 +15,7 @@ class STTNewsroomNinjsFormatter(NewsroomNinjsFormatter):
         self.can_export = False
         self.internal_renditions = ["original", "viewImage", "baseImage"]
 
-    def update_ninjs_stt_sources(self, ninjs):
+    def update_stt_sources(self, ninjs):
         try:
             stt_sources = [
                 subj["name"]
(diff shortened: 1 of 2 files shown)


server/stt/stt_newsroom_ninjs_formatter.py
```diff
@@ -36,10 +37,65 @@ def update_ninjs_stt_sources(self, ninjs):
...
+            logger.error(f"Error occurred when updating stt sources: {str(e)}")
+
+    def update_place(self, ninjs):
+        locators = [place.get("code") for place in ninjs.get("place", [])]
+        if locators:
+            ninjs["locators"] = locators
+
+    def update_body_from_content_profiles(self, ninjs):
+        extra = ninjs.get("extra", {})
+        profile = extra.get("content_profile_name", ninjs.get("profile"))
```
MarkLark86: `content_profile_name` is a special field from the [sttversion CV](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L1095), and not stored anywhere on articles.

Considering the [ContentProfiles](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/content_types.json#L3) CV uses the name for the `_id` (such as `sms` or `viiva`), we can just use `ninjs.get("profile")`

server/stt/stt_newsroom_ninjs_formatter.py
```diff
@@ -36,10 +37,65 @@ def update_ninjs_stt_sources(self, ninjs):
...
+            body = ninjs.get("body_html", "")
+            ninjs["body_html"] = header + (body or "")
+
+    def update_sttversion(self, ninjs):
+        extra = ninjs.get("extra", {})
+        profile = extra.get("content_profile_name", ninjs.get("profile"))
+        if not profile:
+            return
+        versions = str(profile).strip()
+        ninjs["sttversion"] = versions
```
MarkLark86, line 83: There's an implementation note on the ticket, to use the [sttversion CV](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L1095) to get the `content_profile_name` to be used here.
```
Implementation note:
The CV was extended with content_profile_name.
It would be nice to use it in case we will need to add more content profiles in future
```

Therefor we should get the CV (matching CV item's `name.lower()` to `profile.lower()`), then storing that CV item's `content_profile_name` to `sttversion`, falling back to `profile` if CV item not found

server/stt/stt_newsroom_ninjs_formatter.py
```diff
@@ -36,10 +37,65 @@ def update_ninjs_stt_sources(self, ninjs):
 
             ninjs["source"] = "-".join(sorted_sources)
         except Exception as e:
-            logger.error(f"Error occurred when updating ninjs stt sources: {str(e)}")
+            logger.error(f"Error occurred when updating stt sources: {str(e)}")
+
+    def update_place(self, ninjs):
```
MarkLark86: This one won't be needed here, it will be handled in Newshub (Ticket states `place` is already done, STT-1215)

MarkLark86 (commented):
Looks good, just a couple small things

MarkLark86 (approved):

PR #499, opened 2025-10-31
fix date trs_value

minhpham212:
fix  str_date only date no time zone


server/stt/io/feeding_services/stt_tt_content_api.py
```diff
@@ -187,7 +187,7 @@ def _fetch_tt_data(self, provider, update) -> List[Dict]:
             dt_from = datetime.now(timezone.utc) - timedelta(minutes=minutes)
         # TT expects 'trs' as an RFC3339 timestamp in UTC with seconds precision
```
petrjasek, line 190: seems like this comment is no longer valid

server/tests/stt_tt_content_api_test.py
```diff
@@ -395,10 +396,11 @@ def test_fetch_data_adds_trs_with_last_updated(self, mock_get):
...
-        # First call URL should contain trs param with the exact timestamp
+        # First call URL should contain trs param derived from the last_update date
         first_call_args = mock_get.call_args_list[0]
         first_url = first_call_args[0][0]
-        self.assertIn("trs=2025-09-24T10%3A00%3A00Z", first_url)  # URL-encoded ':'
+        qs = parse_qs(urlsplit(first_url).query)
+        self.assertEqual(["2025-09-24"], qs.get("trs"))
 
     @patch.object(STTTTContentAPIService, "_get_with_retry")
     def test_fetch_data_uses_trs_fallback_since_minutes(self, mock_get):
```
petrjasek, line 406: I wonder if that since_minutes option makes sense if you fetch whole day, should be probably `since_days` rather if needed at all

PR #501, opened 2025-11-03
Extend ingest parser for STT planning items [STT-1429]

devketanpro:


server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+        if not has_stt_fields:
+            return
+
+        # Parse headline
+        headline_elt = planning_elt.find(self.qname("headline"))
+        if headline_elt is not None and headline_elt.text:
+            coverage.setdefault("planning", {})["headline"] = headline_elt.text.strip()
+
+        # Parse scheduled/due date
+        workstartdate_elt = planning_elt.find(
```
MarkLark86, line 195: The proper way to do this is using the `qname` function with the `ns` argument. i.e.

https://github.com/superdesk/superdesk-stt/blob/4a929124ddf5f812dfbe3ec4699f38cb5b54ecf2/server/stt/stt_events_ml.py#L25-L27

https://github.com/superdesk/superdesk-stt/blob/4a929124ddf5f812dfbe3ec4699f38cb5b54ecf2/server/stt/stt_events_ml.py#L339-L343

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+    ):
+        """Parse STT-specific coverage fields from XML - only for internal planning items"""
+        planning_elt = news_coverage_elt.find(self.qname("planning"))
+        if planning_elt is None:
+            return
+
+        has_stt_fields = False
+
+        # Check for STT-specific namespaces and qcodes
+        for subj in planning_elt.findall(self.qname("subject")):
```
MarkLark86: What is the benefit in checking for any elements in the document first, and then later re-finding/iterating them?

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+            coverage.setdefault("planning", {})["headline"] = headline_elt.text.strip()
+
+        # Parse scheduled/due date
+        workstartdate_elt = planning_elt.find(
+            ".//{http://www.stt.fi/internal}workstartdate"
+        )
+        if (
+            workstartdate_elt is not None
+            and workstartdate_elt.text
+            and not coverage.get("planning", {}).get("scheduled")
```
MarkLark86: Looks like a Coverage should _always_ have a [scheduled date](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/feed_parsers/superdesk_planning_xml.py#L232).

For STT's use case, I would assume if there is a `workstartdate` defined, we should use that and fallback to what is already populated from core parser.

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+                    "sttworkstatus:1": "ncostat:int",
+                    "sttworkstatus:2": "ncostat:int",
+                    "sttworkstatus:3": "ncostat:int",
+                    "sttworkstatus:4": "ncostat:notint",
+                    "sttworkstatus:5": "ncostat:notdec",
+                }
+
+                mapped_status = status_mapping.get(qcode)
+                if mapped_status:
+                    coverage["news_coverage_status"] = {
```
MarkLark86: We should directly use the [CV stored](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L106) in the DB and managed in the front-end.

This way any changes to this DB won't require code changes to reflect it.

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+        self.parse_coverage_status(planning_elt, coverage)
+        self.parse_coverage_internal_note(planning_elt, coverage)
+        self.parse_picture_type(planning_elt, coverage)
+        self.parse_photographer_awareness(planning_elt, coverage)
+        self.parse_finnish_text_fields(planning_elt, coverage)
+        self.parse_registration_info(planning_elt, coverage)
+
+    def parse_coverage_status(self, planning_elt: Element, coverage: Dict[str, Any]):
+        """Parse coverage status from sttworkstatus subject"""
+        for subject_elt in planning_elt.findall(self.qname("subject")):
```
MarkLark86: Looks like we're iterating the `subject` fields many times, 1 for each type of data we're populating.
An improvement might be to traverse the `subject` fields once, and call the appropriate `parse_` function based on node attributes.

i.e.
```python
for subject_elt in planning_elt.findall(self.qname("subject")):
    qcode = subject_elt.get("qcode", "")
    if qcode.startswith("sttworkstatus:"):
        parse_coverage_status(subject_elt, coverage)
    elif qcode == "sttinternaltext":
        parse_coverage_internal_note(subject_elt, coverage)
    ...
```

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+    def parse_coverage_internal_note(
+        self, planning_elt: Element, coverage: Dict[str, Any]
+    ):
+        """Parse internal note for coverage from sttinternaltext subject"""
+        # Find subject with qcode sttinternaltext
+        for subject_elt in planning_elt.findall(self.qname("subject")):
+            if subject_elt.get("qcode") == "sttinternaltext":
+                value_elt = subject_elt.find(self.qname("value"))
+                if value_elt is not None and value_elt.text:
+                    coverage["internal_note"] = value_elt.text
```
MarkLark86: Like we're doing with the headline, it would be good to strip whitespace from here too

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+                break
+
+    def parse_picture_type(self, planning_elt: Element, coverage: Dict[str, Any]):
+        """Parse picture type from genre and sttimagetypename subject"""
+        # First try genre element
+        genre_elt = planning_elt.find(self.qname("genre"))
+        if genre_elt is not None:
+            qcode = genre_elt.get("qcode")
+            if qcode and qcode.startswith("sttimage:"):
+                coverage.setdefault("planning", {})["sttimagetype"] = qcode
```
MarkLark86: The data is stored into a custom CV on the Coverage, so we should retrieve the item from the [sttimatetype](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L922) CV and stored in the Coverage's `planning.subject` field

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+                break
+
+    def parse_photographer_awareness(
+        self, planning_elt: Element, coverage: Dict[str, Any]
+    ):
+        """Parse photographer awareness from sttphotoaware subject"""
+        for subject_elt in planning_elt.findall(self.qname("subject")):
+            qcode = subject_elt.get("qcode", "")
+            if qcode.startswith("sttphotoaware:"):
```
MarkLark86: This should use the [sttdoesphotographerknow](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L856) CV, and stored in Coverage custom cv.

If the value is `sttphotoaware:0`, then we don't store this value at all

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +136,221 @@ async def get_coverage_details(
...
+                "sttpicturewhatisphotographed"
+            ] = picture_what_is_photographed
+
+    def parse_registration_info(self, planning_elt: Element, coverage: Dict[str, Any]):
+        """Parse registration info from sttentryinfo subject"""
+        for subject_elt in planning_elt.findall(self.qname("subject")):
+            if subject_elt.get("qcode") == "sttentryinfo":
+                value_elt = subject_elt.find(self.qname("value"))
+                if value_elt is not None and value_elt.text:
+                    coverage.setdefault("planning", {})[
```
MarkLark86: Again, this is a custom text field, stored in `coverages.planning.fields`

server/stt/stt_planning_ml.py
```diff
@@ -251,19 +477,26 @@ def get_coverage_type(coverage):
...
+                "planning": {
+                    "slugline": "",
+                    "g2_content_type": "text",
+                    "scheduled": item.get("planning_date"),
+                },
+                "flags": {"placeholder": True},
+            }
+
+            # Set default coverage status for placeholder
+            placeholder_coverage["news_coverage_status"] = {
```
MarkLark86: We should really be using data directly from the CV, in case of any changes to the CV label/name and translations

MarkLark86 (changes requested):
In general looks fine, mostly invalid storage location for custom CV and text fields

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +139,227 @@ async def get_coverage_details(
...
+            value_text = (
+                value_elt.text.strip()
+                if value_elt is not None and value_elt.text
+                else ""
+            )
+
+            if qcode.startswith("sttworkstatus:"):
+                self.parse_coverage_status(subject_elt, coverage)
+            elif qcode == "sttinternaltext":
+                coverage["internal_note"] = value_text
```
MarkLark86: The `internal_note` field is in the `planning` dictionary

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +139,227 @@ async def get_coverage_details(
...
+        self, subject_elt: Element, coverage: Dict[str, Any]
+    ):
+        """Parse photographer awareness from sttphotoaware subject"""
+        qcode = subject_elt.get("qcode", "")
+
+        # Only store if photographer knows (sttphotoaware:2) or doesn't know (sttphotoaware:1)
+        # Don't store for sttphotoaware:0 (not set)
+        if qcode in ["sttphotoaware:1", "sttphotoaware:2"]:
+            photographer_awareness = self.get_photographer_awareness_from_vocabulary(
+                qcode
```
MarkLark86: [qcode](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L875) is either `yes` or `no`

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +139,227 @@ async def get_coverage_details(
...
+            if qcode.startswith("sttworkstatus:"):
+                self.parse_coverage_status(subject_elt, coverage)
+            elif qcode == "sttinternaltext":
+                coverage["internal_note"] = value_text
+            elif qcode.startswith("sttimagetypename:"):
+                self.parse_picture_type(subject_elt, coverage)
+            elif qcode.startswith("sttphotoaware:"):
+                self.parse_photographer_awareness(subject_elt, coverage)
+            elif qcode == "sttentryinfo":
+                coverage["planning"]["fields"]["sttregistrationinfo"] = value_text
```
MarkLark86: Please check the schema for [fields](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning_schema.py#L134), as it is not a dictionary

MarkLark86 (changes requested):
Still an issue with the Coverage custom text fields (stored in `planning.fields`).
Also, can you please update the [sttimagetype](https://github.com/superdesk/superdesk-stt/blob/develop/server/data/vocabularies.json#L935) with the CV provided in the ticket

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +139,253 @@ async def get_coverage_details(
...
+            )
+
+        return None
+
+    def parse_picture_type(self, subject_elt: Element, coverage: Dict[str, Any]):
+        """Parse picture type from sttimagetypename subject"""
+        qcode = subject_elt.get("qcode", "")
+
+        # Map STT numeric image type codes to new CV qcodes
+        image_type_mapping = {
```
MarkLark86: The CV provided in the ticket, which you have to update the data/vocabularies.json file with, uses the `sttimage:20` as qcode.
Please update the data/vocabularies.json file, and update this qcode mapping

server/stt/stt_planning_ml.py
```diff
@@ -124,7 +139,253 @@ async def get_coverage_details(
...
+            "sttimagetypename:20": "sttimage:20",  # Kuvaaja paikalla
```
MarkLark86: Where is this mapping coming from, I couldn't find it?

devketanpro: This mapping comes from the updated sttimagetype vocabulary defined in `server/data/vocabularies.json` and `STT_Imagetype.json.`

MarkLark86: I mean the number mapping,
sttimagetypename to sttimage qcode?
Like this one
"sttimagetypename:29": "sttimage:70",  # Video

Where is 29 mapped to 70, I couldn't find that

devketanpro: Sample files

the mapping was derived from analysing the sample XML files (`690975_timefix.xml`,` 691631_timefix.xml`) provided in the ticket, where you can see which sttimagetypename values correspond to which image types in the actual data.

MarkLark86 (approved):

PR #531, opened 2025-11-11
Extend ingest parser for STT planning items [STT-1429]

devketanpro:
added custom vocabulary and updated parser to map scema

### DIFF
--- a/server/data/vocabularies.json
+++ b/server/data/vocabularies.json
@@ -934,7 +934,7 @@
     {
         "_id": "sttimagetype",
         "init_version": 1,
-        "field_type": "null",
+        "field_type": null,
         "items": [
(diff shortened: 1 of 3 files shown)


server/data/vocabularies.json
```diff
@@ -933,8 +933,8 @@
     },
     {
         "_id": "sttimagetype",
-        "init_version": 1,
-        "field_type": null,
+        "init_version": 2,
+        "field_type": "custom_vocabulary",
```
petrjasek: we don't set it for vocabularies

petrjasek (commented):
pls add test for those changes

petrjasek (approved):

PR #576, opened 2025-11-25
[STT-1481] - STT NewsML parser ignores hyperlinks

BrianMwangi21:
### Purpose
- This PR ensures STT NewsML ingest preserves the hyperlinks using the core html cleaner function but with an updated config to include <a> tag

### What Has Changed
- `HTML_TAGS_WHITELIST` in the settings.py now includes the <a> tag
- Added tests

### DIFF
--- a/server/stt/parser.py
+++ b/server/stt/parser.py
@@ -59,6 +59,10 @@ class STTParser(STTParserMixin, STTNewsMLFeedParser):
         return content
 
     def _clean_body_html(self, body_elt):
+        """
+        This function retains the core’s cleaner behaviour while forcing <a> tags onto the allow-list
+        for html tags so ingesting doesn't ignore the hyperlinks
+        """
         if body_elt is None:
             return body_elt
 
--- a/server/stt/parser.py
+++ b/server/stt/parser.py
@@ -1,5 +1,4 @@
 from superdesk import etree as sd_etree
-from superdesk.core import get_app_config
 from superdesk.io.registry import register_feed_parser
 from superdesk.io.feed_parsers.stt_newsml import STTNewsMLFeedParser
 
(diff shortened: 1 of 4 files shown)


server/stt/parser.py
```diff
@@ -57,6 +58,32 @@ def parse_inline_content(self, tree, item):
...
+    def _clean_body_html(self, body_elt):
+        """
+        This function retains the core’s cleaner behaviour while forcing <a> tags onto the allow-list
+        for html tags so ingesting doesn't ignore the hyperlinks
+        """
+        if body_elt is None:
+            return body_elt
+
+        allow_tags = list(get_app_config("HTML_TAGS_WHITELIST"))
+        if "a" not in allow_tags:
```
petrjasek: I wouldn't do it here, if there is a config we should follow. you can add `a` to `HTML_TAGS_WHITELIST` in `settings.py`

server/stt/parser.py
```diff
@@ -57,6 +58,32 @@ def parse_inline_content(self, tree, item):
...
+            allow_tags.append("a")
+
+        if not isinstance(body_elt, sd_etree.html.HtmlElement):
```
petrjasek: I wonder when would this happen?

server/stt/parser.py
BrianMwangi21: I had thought of this but I thought since it didn't exist in this repos `settings.py` that I could localize it to this function since the whitelist is used by clean_html and that is used by other functions and could have unintended effect, like now the <a> link showing in other parts where it isn't intended. 

Does that make sense ?

server/stt/parser.py
BrianMwangi21: This one is there as a fallback cause the core also has a similar check -> https://github.com/superdesk/superdesk-core/blob/9d38e4fb15a702589d2ea0e7decf97155cd24abd/superdesk/etree.py#L173

server/stt/parser.py
petrjasek: in that case I would rather not use the config and just put the full list here

server/stt/parser.py
petrjasek: actually I'm wondering why not use that core function here with updated config

petrjasek (approved):

PR #583, opened 2025-11-27
Updated package-lock.json

BrianMwangi21:


petrjasek:
pls don't do this manually, it's always doing more changes than needed

PR #592, opened 2025-12-02
fix stt newsml formatter errors from sentry

petrjasek:
- fix error when ednote is None
- fix error when subheadline has additional markup

STT-1497

### DIFF
--- a/server/stt/publish/stt_newsml_g2.py
+++ b/server/stt/publish/stt_newsml_g2.py
@@ -426,12 +426,12 @@ class STTNewsmLG2Formatter(NewsMLG2Formatter):
 
             edNoteStr = newsroomnoteStr + (article.get("ednote") or "")
             SubElement(itemMeta, "edNote", attrib={"role": "sttnote:private"}).text = (
-                edNoteStr
+                edNoteStr.strip()
             )
         else:
             edNoteStr = article.get("ednote", "")
             SubElement(itemMeta, "edNote", attrib={"role": "sttnote:private"}).text = (
-                edNoteStr
+                edNoteStr.strip()
             )
 
         # Public edNote - to all other profiles but 'nettiuutinen'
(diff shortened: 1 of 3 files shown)


server/stt/publish/stt_newsml_g2.py
```diff
@@ -623,8 +623,13 @@ def format_contentSet(self, article, parentNode):
                 subheadline = etree.Element("h2")
 
                 # Find the paragraph and get the content to H2
-                subheadline.text = htmltree.xpath("//p/text()")[0]
-                body.append(subheadline)
+                try:
+                    subheadline_p = htmltree.xpath("//p")[0]
+                except IndexError:
+                    subheadline_p = None
+                if subheadline_p is not None and subheadline_p.text_content():
```
eos87, line 630: I'm wondering if we should consider the possibility of `.text_content` not being available. I've seen that fail in the past for some odd parsing reasons.

server/stt/publish/stt_newsml_g2.py
```diff
@@ -424,14 +424,14 @@ def format_itemMeta(self, article, parentNode):
             else:
                 newsroomnoteStr = newsroomnoteStr + ". "
 
-            edNoteStr = newsroomnoteStr + article.get("ednote", "")
+            edNoteStr = newsroomnoteStr + (article.get("ednote") or "")
```
eos87, line 427: what are the odds of `article.get("ednote")` returning a non-string value? like if this would be an integer the concatenation would throw an error

server/stt/publish/stt_newsml_g2.py
petrjasek: we have a control on the markup in the field, there is always `p` inside if it's present

server/stt/publish/stt_newsml_g2.py
petrjasek: 0%, it's string or `None`

eos87 (approved):

PR #596, opened 2025-12-03
[STT-1484] - Use new config to set `send_to` as default authoring tab for stt

BrianMwangi21:


server/settings.py
```diff
@@ -222,6 +222,9 @@
 
 # AUTHORING CONFIG
 
+# set send_to tab as the default opened tab in authoring actions
+AUTHORING_ACTIONS_DEFAULT_TAB = env("AUTHORING_ACTIONS_DEFAULT_TAB", "send_to")
```
petrjasek: imo no need to use env here

petrjasek (approved):

PR #606, opened 2025-12-08
Fix STT SMS formatter ndash issue

msilven:
Fixes STT's specific ndash and thinsp notations in STT SMS formatter.


petrjasek:
hi @msilven I've done some cleanup before so I've created a new PR with those latest changes to avoid conflics
https://github.com/superdesk/superdesk-stt/pull/607

PR #614, opened 2025-12-10
tweak Procfile

petrjasek:
- use specific env vars for capi/papi workers
- use redbeat for celery

STT-1523


MarkLark86 (approved):

MarkLark86:
Looks good. I would suggest for a future improvement we could place the hypercorn_config.py in superdesk-core, and import that from each app repo (similar to how we do it for grunt).

And we could make the default config to apply these worker settings.
i.e.
```python
import os
import multiprocessing
import sys

bind = "0.0.0.0:%s" % os.environ.get("PORT", "5000")

worker_env_name: str
default_workers: int 
if "content_api.wsgi:application" in sys.argv:
    worker_env_name = "CAPI_WORKERS"
    default_workers = 2
elif "prod_api.wsgi:application" in sys.argv:
    worker_env_name = "PAPI_WORKERS"
    default_workers = 2
else:
    worker_env_name = "WEB_CONCURRENCY"
    default_workers = multiprocessing.cpu_count() + 1

workers = int(os.environ.get(worker_env_name, default_workers))

accesslog = "-"
access_log_format = "%(m)s %(U)s status=%(s)s time=%(T)ss size=%(B)sb"

use_reloader = "SUPERDESK_RELOAD" in os.environ
timeout = int(os.environ.get("WEB_TIMEOUT", 30))
```

Thoughts?

petrjasek:
yep that would simplify it. not sure if that default cpu count + 1 makes sense for hypercorn, I think that was the default for gunicorn

PR #734, opened 2026-03-04
STT-1584: Add tooltips with keyboard shortcuts to custom editor tags

BrianMwangi21:


MarkLark86:
@BrianMwangi21 ~~This should rather go into the `develop` branch (STT1.4), as `next` branch is being used for STT1.3~~ Forget that.

I assume this doesn't require any core code changes, it's merely a config. If that's the case then we may be fine squeezing that one into the STT1.3 release

BrianMwangi21:
Actually there is a client-core code change here -> https://github.com/superdesk/superdesk-client-core/pull/5100

Waiting on Helmy to test it out.

eos87 (approved):

PR #736, opened 2026-03-05
fix: align server-side HTML character count with client logic

0is1:
Update the `count_body_html_characters` Jinja filter to match the logic used by the client-side `cleanHtml` and `getEditorFieldCharactersCount` utilities.

Previously, the server-side function only unescaped HTML entities without stripping the actual HTML tags or embeds. This caused significant discrepancies in the reported character counts on the server (e.g., reporting 4400 characters for an article that visibly only has 2900 characters).

The new implementation now correctly evaluates the text length by:
- Stripping out `<!-- EMBED START ... -->` wrapper blocks completely.
- Translating `<br>` variations into spaces.
- Stripping out all remaining HTML tags.
- Removing invisible line breaks (`\n` and `\r`).
- Unescaping HTML entities (like `&amp;`) before the final count, so they are accurately counted as single characters.
- Adding a comprehensive test suite in `server/tests/template_filters_test.py` to cover all formatting transformations.


MarkLark86:
@0is1 There is [get_word_count](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/text_utils.py#L80) in superdesk-core that does something similar to this PR, and should be kept in sync with the front-end.

If there are any changes you'd like, you can create a PR for changes in [get_text_word_count](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/text_utils.py#L24-L55) function, which uses regex in a similar way that you have here.

0is1:
@MarkLark86 yes but this is about counting characters that is totally different thing. I haven't found "count characters" helper from the backend core but if there's that kind of helper already, let me know. This is trying to behave same way as `getEditorFieldCharactersCount` in the client https://github.com/superdesk/superdesk-client-core/blob/e9dd6d4623465d8050acfd0aaa5b3ca70aa382b6/scripts/apps/authoring/authoring/components/CharacterCount.tsx#L13-L20

MarkLark86:
> @MarkLark86 yes but this is about counting characters that is totally different thing. I haven't found "count characters" helper from the backend core but if there's that kind of helper already, let me know. This is trying to behave same way as `getEditorFieldCharactersCount` in the client https://github.com/superdesk/superdesk-client-core/blob/e9dd6d4623465d8050acfd0aaa5b3ca70aa382b6/scripts/apps/authoring/authoring/components/CharacterCount.tsx#L13-L20

Oh right, my bad.

In that case, it would be good to reuse some of that same code from core, that uses a library to remove html entities.

0is1:
> > @MarkLark86 yes but this is about counting characters that is totally different thing. I haven't found "count characters" helper from the backend core but if there's that kind of helper already, let me know. This is trying to behave same way as `getEditorFieldCharactersCount` in the client https://github.com/superdesk/superdesk-client-core/blob/e9dd6d4623465d8050acfd0aaa5b3ca70aa382b6/scripts/apps/authoring/authoring/components/CharacterCount.tsx#L13-L20
> 
> Oh right, my bad.
> 
> In that case, it would be good to reuse some of that same code from core, that uses a library to remove html entities.

Is there that kind of logic in the backend core somewhere? Because that client code is typescript and backend uses python. So that isn't reusable as is in this case.

MarkLark86:
> > > @MarkLark86 yes but this is about counting characters that is totally different thing. I haven't found "count characters" helper from the backend core but if there's that kind of helper already, let me know. This is trying to behave same way as `getEditorFieldCharactersCount` in the client https://github.com/superdesk/superdesk-client-core/blob/e9dd6d4623465d8050acfd0aaa5b3ca70aa382b6/scripts/apps/authoring/authoring/components/CharacterCount.tsx#L13-L20
> > 
> > Oh right, my bad.
> > 
> > In that case, it would be good to reuse some of that same code from core, that uses a library to remove html entities.
> 
> Is there that kind of logic in the backend core somewhere? Because that client code is typescript and backend uses python. So that isn't reusable as is in this case.

Sorry, I was referring to backend core code:

https://github.com/superdesk/superdesk-core/blob/develop/superdesk%2Ftext_utils.py#L110

0is1:
> > > > @MarkLark86 yes but this is about counting characters that is totally different thing. I haven't found "count characters" helper from the backend core but if there's that kind of helper already, let me know. This is trying to behave same way as `getEditorFieldCharactersCount` in the client https://github.com/superdesk/superdesk-client-core/blob/e9dd6d4623465d8050acfd0aaa5b3ca70aa382b6/scripts/apps/authoring/authoring/components/CharacterCount.tsx#L13-L20
> > > 
> > > 
> > > Oh right, my bad.
> > > In that case, it would be good to reuse some of that same code from core, that uses a library to remove html entities.
> > 
> > 
> > Is there that kind of logic in the backend core somewhere? Because that client code is typescript and backend uses python. So that isn't reusable as is in this case.
> 
> Sorry, I was referring to backend core code:
> 
> https://github.com/superdesk/superdesk-core/blob/develop/superdesk%2Ftext_utils.py#L110

Okay, thanks for pointing that out 👍

MarkLark86 (approved):

PR #740, opened 2026-03-06
feat: Improve Newshub search filters and normalize fetched search items

0is1:
Replace the old Newshub filter wiring with explicit single-select controls for service, genre, version type and urgency, and add a free-text byline field in the custom search panel.

Update the Newshub search provider to:
- request a richer field set via include_fields
- map client filters to the API fields used by news/search (categories -> service, sttversion -> subject, urgency, genre)
- append byline text to the free-text query (it seems that `/news/search` endpoint doesn't support `byline` / `author` search param directly, it doesn't also return that `byline` / `author` data)
- normalize search results as externalsource text items so built-in Fetch and Fetch To actions work correctly
- fetch items by _id search and return the first matching result

### DIFF
--- a/server/stt/search_providers/newshub.py
+++ b/server/stt/search_providers/newshub.py
@@ -5,6 +5,7 @@ from datetime import datetime, timezone
 import arrow
 import aiohttp
 from aiohttp.client_exceptions import ClientResponseError
+from dateutil import parser as dtparse
 
 import superdesk
 from superdesk.core import get_config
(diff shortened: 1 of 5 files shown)


server/stt/search_providers/newshub.py
```diff
@@ -27,7 +29,7 @@ def count(self, **kwargs) -> int:
 class NewshubSearchProvider(superdesk.SearchProvider):
     label = "Newshub"
     # TODO: set the base_url to the production URL when ready
-    base_url = "https://stt-uat.newshub.pro/newsapi/v1/"
+    base_url = "https://stt-next.newshub.pro/newsapi/v1/"
```
petrjasek, line 32: might be good to use an env variable for it

petrjasek (approved):

